PR #4, opened 2022-02-18
[TOAZ-67] Enables Jupyter Lab access by supporting msg fragmentation for relayed messages

giventocode:
- Text messages are relayed using message fragmentation when target requires it.
  - Fixes issue with Jupyter Lab notebooks. 
- Refactoring and additional testing.


Qi77Qi:
tested locally...things seem to be working

Do you think we should update https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/main/service/src/main/java/org/broadinstitute/listener/relay/wss/ConnectionsPair.java#L37 as well? altho it's not obvious to me where we would get that `last` param...if we can't figure it out immediately, maybe add a TODO comment just a reminder?

giventocode:
> tested locally...things seem to be working
> 
> Do you think we should update https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/main/service/src/main/java/org/broadinstitute/listener/relay/wss/ConnectionsPair.java#L37 as well? altho it's not obvious to me where we would get that `last` param...if we can't figure it out immediately, maybe add a TODO comment just a reminder?

Yep and I looked into it. I think we are fine. Internally, in the SDK, the read operation dequeues the message from a `textQueue` and the `textQueue` is populated by the jetty websocket listener `onMessage(String)` method - [details here](https://github.com/Azure/azure-relay-java/blob/2add12f85cfc5ab08ed1422d6d96bffddc8eb94b/src/main/java/com/microsoft/azure/relay/ClientWebSocket.java#L329)

Now, according to the jetty docs, this method is called when all the fragments have been received and the message is complete. More details [here](https://archive.eclipse.org/jetty/8.1.0.RC5/apidocs/org/eclipse/jetty/websocket/WebSocket.OnTextMessage.html)

Qi77Qi:
@giventocode oh awesome....I can maybe test out if things work as expected when running a large cell...

PR #8, opened 2022-04-12
TOAZ-74 Add Sam check

Qi77Qi:
https://broadworkbench.atlassian.net/browse/TOAZ-74

* Add sam check
* Upgrade to java 17
* Add test run in CI


service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -0,0 +1,68 @@
...
+    var cookieValue = headers.getOrDefault("cookie", headers.get("Cookie"));
+    String[] splitted = cookieValue.split(";");
+
+    var leoToken = Arrays.stream(splitted)
+        .takeWhile(s -> s.contains("LeoToken"))
+        .findFirst()
+        .map(s -> s.split("=")[1]);
+
+    if (leoToken.isEmpty()) {
+      logger.info("No valid cookie found " + cookieValue);
```
giventocode, line 53: Are we okay logging the cookie value?  Even if the Leo token is empty, it may contain sensitive information.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -0,0 +1,68 @@
...
+      return false;
+    } else {
+      var token = leoToken.get();
+      samClient.setAccessToken(token);
+      samClient.setBasePath(listenerProperties.getSamUrl());
+      var resourceApi = new ResourcesApi(samClient);
+      try {
+        return resourceApi.resourcePermissionV2(SAM_RESOURCE_TYPE, listenerProperties.getSamResourceId(), "write");
```
giventocode, line 61: Should be consider caching, looking at a similar implementation in another service?  If the token has an expiration timestamp, then the check should be valid for that duration. The implementation must be thread-safe as well.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -0,0 +1,68 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.lang.NonNull;
+import org.springframework.stereotype.Component;
+
+@Component(InspectorNameConstants.SAM_CHECKER)
+public class SamPermissionInspector implements RequestInspector {
+  private final Logger logger = LoggerFactory.getLogger(SamPermissionInspector.class);
+  private final String SAM_RESOURCE_TYPE = "controlled-application-private-workspace-resource";
+  private final ApiClient samClient = new ApiClient();
```
giventocode, line 21: maybe this client should be injected not instantiated in the class.

service/src/main/java/org/broadinstitute/listener/config/ListenerProperties.java
```diff
@@ -20,6 +22,22 @@ public void setRelayConnectionString(String relayConnectionString) {
...
+
+  public void setSamUrl(String samUrl) {
+    this.samUrl = samUrl;
+  }
+
+  public String getSamResourceId() {
+    return samResourceId;
+  }
+
+  public void setSamResourceId(String samResourceId) {
```
giventocode, line 37: Will this properties be required if the inspector is not enabled?

giventocode (changes requested):
Good start and makes sense. Left some comments.

service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
```diff
@@ -0,0 +1,24 @@
+package org.broadinstitute.listener.config;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+public class SamInspectorProperties {
```
pshapiro4broad, line 5: This seems like a good place to use `record`, assuming that the setters aren't used and the fields are actually `final`.

service/src/main/java/org/broadinstitute/listener/config/ListenerProperties.java
```diff
@@ -10,6 +10,16 @@ public class ListenerProperties {
   private String relayConnectionString;
```
pshapiro4broad, line 10: This class could probably be converted to a record; you can have records populated by `@ConfigurationProperties`.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -0,0 +1,72 @@
...
+  protected Optional<String> getToken(@NotNull String cookieValue) {
+    String[] splitted = cookieValue.split(";");
+
+    if(cookieValue.isEmpty())
+      return Optional.empty();
+    else
+      return Arrays.stream(splitted)
+        .filter(s -> s.contains("LeoToken"))
+        .findFirst()
+        .map(s -> s.split("=")[1]);
```
pshapiro4broad, line 70: Google java code style (which all terra follows) requires brackets around all statements. So this needs some `{`s.

Although in this case the `if` is unnecessary. You can use `Arrays.stream(splitted)` even if `cookieValue.isEmpty()`.
```suggestion
      return Arrays.stream(splitted)
        .filter(s -> s.contains("LeoToken"))
        .findFirst()
        .map(s -> s.split("=")[1]);
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -0,0 +1,61 @@
...
+    this.tokenChecker = tokenChecker;
+    this.samClient = samClient;
+  }
+
+  // Should only be used in checkCachedPermission, but making it public so that we can test it
+  @Cacheable("expiresAt")
+  public Instant checkWritePermission(String accessToken) {
+    try {
+      var oauthInfo = tokenChecker.getOauthInfo(accessToken);
+      if(oauthInfo.expiresAt().isPresent()) {
```
pshapiro4broad, line 32: Google code style also requires a space between `if` and `(`. This is to help visually distinguish function calls from non-function calls.
```suggestion
      if (oauthInfo.expiresAt().isPresent()) {
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
Qi77Qi: >Google java code style (which all terra follows) requires brackets around all statements.
dang...I don't wanna write ";" if java allows it

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
Qi77Qi: I'll leave all the formatting to spotless, which is currently broken due to java 17 upgrade...think we can leave that out for this PR

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -0,0 +1,61 @@
...
+        samClient.setAccessToken(accessToken);
+        var resourceApi = new ResourcesApi(samClient);
+
+        var res = resourceApi.resourcePermissionV2(SAM_RESOURCE_TYPE, samResourceId, "write");
+        if(res)
+          return oauthInfo.expiresAt().get();
```
pshapiro4broad, line 38: This is totally a code style issue, but I would write this so that all cases fall through to the end of the method except for this one. They all return the same value, so only one `return` is needed. IMO this makes it clearer that all returns are the same.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
Qi77Qi: appreciate it if you have some idea how to fix spotless tho

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -0,0 +1,61 @@
...
+      } else {
+        logger.error("Token expired " + oauthInfo.error());
+        return Instant.EPOCH;
+      }
+    } catch (IOException e) {
+      logger.error("Fail to check token info", e);
+      return Instant.EPOCH;
+    } catch (InterruptedException e) {
+      logger.error("Fail to check token info", e);
+      return Instant.EPOCH;
```
pshapiro4broad, line 50: Java lets you combine multiple `catch` clauses that have the same body. IntelliJ can do this fix for you automatically.
```suggestion
    } catch (IOException | InterruptedException  e) {
      logger.error("Fail to check token info", e);
      return Instant.EPOCH;
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -0,0 +1,61 @@
...
+  private final Logger logger = LoggerFactory.getLogger(SamResourceClient.class);
+
+  public OauthInfo getOauthInfo(String token) throws IOException, InterruptedException {
+    var jwtExpiration = tryDecodeAsB2cToken(token);
+
+    var now = Instant.now();
+
+    if (jwtExpiration.isPresent()) {
+      if(jwtExpiration.get().isAfter(now))
+        return new OauthInfo(Optional.of(jwtExpiration.get()), "");
```
pshapiro4broad, line 27: this is redundant, I think
```suggestion
        return new OauthInfo(jwtExpiration, "");
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -0,0 +1,61 @@
...
+      var decoded = new Gson().fromJson(oauthInfoResponse.body(), GoogleOauthInfoResponse.class);
+
+      if(decoded.expires_in > 0)
+          return new OauthInfo(Optional.of(now.plusSeconds(decoded.expires_in)), decoded.error);
+      else
+        return new OauthInfo(Optional.empty(), decoded.error);
+    }
+  }
+
+  private Optional<Instant> tryDecodeAsB2cToken(String token) {
```
pshapiro4broad, line 45: In this case I don't think `Optional` is buying you much. I would change the code to return `null` directly, since there's no valid result that could be `null`.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+  }
+}
+
+class GoogleOauthInfoResponse {
+  int expires_in;
+  String error;
+}
+
+record OauthInfo(Optional<Instant> expiresAt, String error) {}
```
pshapiro4broad, line 61: Since there's no use of this class yet, it's hard to tell for sure, but I'd guess that using a nullable `Instant` directly here instead of `Optional` would result in less code. Also typically in java you don't have more than one class in a .java file. If you have a "helper" class it's usually declared as an inner class.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -0,0 +1,61 @@
...
+        return new OauthInfo(Optional.of(jwtExpiration.get()), "");
+      else
+        return new OauthInfo(Optional.empty(), "JWT expired");
+    } else {
+      var request = HttpRequest.newBuilder()
+          .uri(URI.create(GOOGLE_OAUTH_SERVER + token))
+          .build();
+
+      var oauthInfoResponse = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+      var decoded = new Gson().fromJson(oauthInfoResponse.body(), GoogleOauthInfoResponse.class);
```
pshapiro4broad, line 36: For new java code I would recommend Jackson over Json, since spring uses Jackson and we use spring boot for our services. The equivalent code would be
```
      var decoded = new ObjectMapper().readValue(oauthInfoResponse.body(), GoogleOauthInfoResponse.class);
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -0,0 +1,61 @@
...
+      return Optional.of(decoded.getExpiresAt().toInstant());
+    } catch (com.auth0.jwt.exceptions.JWTDecodeException e) {
+      logger.info("Fail to check decode JWT", e);
+      return Optional.empty();
+    }
+  }
+}
+
+class GoogleOauthInfoResponse {
+  int expires_in;
```
pshapiro4broad, line 57: Even though this is mapped from a JSON object, I think camel case would work here, as Jackson (and Gson) map between snake and camel case when mapping JSON to a java object. A record could work here too, I think.
```suggestion
  int expiresIn;
```

pshapiro4broad (commented):
Hi, just a few comments

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
Qi77Qi: in Scala, I totally would..but how would you do that in java

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
Qi77Qi: think I'll stick to `Optional`....if later `Optional` becomes a bit more powerful, then the call site can be simplied as well...would prefer it if I don't have to do null check, which is hard to tell if it's from code we write or from somewhere

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
Qi77Qi: I'm gonna stick with `expires_in` since this class is really only meant for the api response from Google..it shouldn't be used anywhere else...I'd rather know what exactly Google returns instead of having to guess there's conversion happened by the json library

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
pshapiro4broad: Something like
```
  public Instant checkWritePermission(String accessToken) {
    try {
      var oauthInfo = tokenChecker.getOauthInfo(accessToken);
      if (oauthInfo.expiresAt() != null) {

        if (resourceApi.resourcePermissionV2(SAM_RESOURCE_TYPE, samResourceId, "write")) {
          return oauthInfo.expiresAt();
        }
      } else {
        logger.error("Token expired " + oauthInfo.error());
      }
    } catch (IOException | InterruptedException e) {
      logger.error("Fail to check token info", e);
    } catch (ApiException e) {
      logger.error("Fail to check Sam permission", e);
    } catch (Exception e) {
      logger.error("Fail for unknown reasons", e);
    }
    return Instant.EPOCH;
  }
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
pshapiro4broad: I have a fix for the spotless config in a branch off of this branch, but can't commit to this repo. If you give me write privs, I'll push the branch to PR into this one.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
pshapiro4broad: OK. In any event I think it'd be better to declare it as a private inner class of `TokenChecker`, since no other class uses it.
```
  static private class GoogleOauthInfoResponse {
    int expires_in;
    String error;
  }
```

service/src/test/java/org/broadinstitute/listener/relay/inspectors/SamResourceClientTest.java
```diff
@@ -0,0 +1,57 @@
...
+  void setUp() throws IOException, InterruptedException, ApiException {
+    samResourceClient = new SamResourceClient("resourceId", apiClient, tokenChecker);
+  }
+
+  @Test
+  void checkWritePermission_sucess() throws IOException, InterruptedException, ApiException {
+    var expiresAt = Instant.now().plusSeconds(100);
+    var oauthResponse = new OauthInfo(Optional.of(expiresAt), "");
+    when(tokenChecker.getOauthInfo(any())).thenReturn(oauthResponse);
+    var apiResponse = new ApiResponse(200, Map.of(), true);
```
pshapiro4broad, line 39: Using a parameterized type like `ApiResponse` without a type argument will generate a warning (and probably sonar will complain too). Looking at this code, `Object` should work:
```suggestion
    var apiResponse = new ApiResponse<Object>(200, Map.of(), true);
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
Qi77Qi: cant make it private for some reason...will leave it as is

pshapiro4broad: That's odd, I tried this out on your branch and it worked fine for me. You can use `record` for this too:
```
  private record GoogleOauthInfoResponse(int expires_in, String error, String error_description) {}
```

service/src/main/resources/application.yml
```diff
@@ -3,6 +3,9 @@ server:
 
 spring:
   application.name: terra-azure-relay-listener
+  cache:
+    cache-names: expiresAt
+    caffeine.spec: maximumSize=100,expireAfterAccess=3600s
```
giventocode, line 8: Is the token expiration rolling (the same token can be renewed/extended when calling the google API) or fixed (user must get another token when one expires)? If the former, then we should adjust the expiration to be equal, or slightly less than the max duration of the token.  Otherwise, tokens will be considered expired when in fact they are not.

giventocode (changes requested):

service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
```diff
@@ -30,6 +35,17 @@ public TargetResolver targetResolver() {
     return new DefaultTargetResolver(properties);
   }
 
+  @Bean
+  public SamResourceClient samResourceClient() {
```
pshapiro4broad, line 39: Is `SamResourceClient` thread safe? I think this bean will have singleton scope by default; if so it will be shared across all threads that inject it.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -0,0 +1,69 @@
...
+      return checkCachedPermission(token);
+    }
+  }
+
+  public boolean checkCachedPermission(String accessToken) {
+    var expiresAt = samResourceClient.checkWritePermission(accessToken);
+    return expiresAt.isAfter(Instant.now());
+  }
+
+  protected Optional<String> getToken(@NotNull String cookieValue) {
```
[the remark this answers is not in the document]
pshapiro4broad: I'm confused by the use of `@NotNull` in `@NotNull String cookieValue` as it looks like it could be `null`, if neither `cookie` or `Cookie` are in the http request header list. Not sure why the not-null checker, if there is one, doesn't produce a warning here.

service/src/main/resources/application.yml
Qi77Qi: it's fixed...Terra UI will peridically keep the token refreshed I think

giventocode (approved):

service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
Qi77Qi: Idk sprint boot well... any comments @giventocode 

will merge this PR for now and address retrospectively if update is needed

PR #10, opened 2022-05-05
IA-3120 Allow requests that has proper Authorization header

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3120

I also reenabled spotless in this PR, so there're some pure formatting diff

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
@@ -34,15 +34,12 @@ public class SamPermissionInspector implements RequestInspector {
   }
 
   private boolean checkPermission(Map<String, String> headers) {
-
     if (headers == null) {
       logger.error("No auth headers found");
       return false;
     }
 
-    var cookieValue = headers.getOrDefault("cookie", headers.get("Cookie"));
-
-    var leoToken = getToken(cookieValue);
+    var leoToken = getToken(headers);
 
     if (leoToken.isEmpty()) {
       logger.error("No valid cookie found");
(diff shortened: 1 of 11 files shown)


service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -47,7 +47,7 @@ private Optional<Instant> tryDecodeAsB2cToken(String token) {
       var decoded = JWT.decode(token);
       return Optional.of(decoded.getExpiresAt().toInstant());
     } catch (com.auth0.jwt.exceptions.JWTDecodeException e) {
-      logger.info("Fail to check decode JWT", e);
+      logger.debug("Fail to check decode JWT", e);
```
giventocode, line 50: Shouldn't this be logger.error?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
```diff
@@ -58,12 +54,23 @@ public boolean checkCachedPermission(String accessToken) {
     return expiresAt.isAfter(Instant.now());
   }
 
-  protected Optional<String> getToken(String cookieValue) {
+  protected Optional<String> getToken(Map<String, String> headers) {
+    return getTokenFromCookie(headers).or(() -> getTokenFromAuthorization(headers));
```
giventocode, line 58: This means the check order would be cookie and then Auth header. I think the reverse may be better. The Auth header is the standard way of checking out. The cookie should be the fallback.  How does Leo currently do it?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -30,9 +30,11 @@ public Instant checkWritePermission(String accessToken) {
     try {
       var oauthInfo = tokenChecker.getOauthInfo(accessToken);
       if(oauthInfo.expiresAt().isPresent()) {
+
         samClient.setAccessToken(accessToken);
         var resourceApi = new ResourcesApi(samClient);
 
+        logger.info("token "+accessToken);
```
giventocode, line 37: Do we want to log the access token? That is sensitive information.

giventocode (changes requested):

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
Qi77Qi: I'm testing things out still...making it debug becuz I'm testing with google token..just to reduce some noise

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
Qi77Qi: we don't..it's just for debugging cuz for some reason things are not working as I'm testing e2e

PR #12, opened 2022-05-13
[TOAZ-140] Fix: Return a HTTP 500 response when requests to the target fail. 

giventocode:
Issue Description: When a request to the target endpoint fails, the listener must send a 500 response to the caller. Instead, the caller is receiving gateway time out error from Azure Relay. 

Root cause: The error handling of the listener is not properly closing the relayed response body when target endpoint fails to be called. The relayed response is not closed because the body does not contain any content. Azure Relay eventually times out the request, and the caller receives a timeout error instead of a 500 response. 
 

Resolution: Add content to the response body so the relayed response is close properly.  Make sure the target response is always closed.

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
@@ -1,9 +1,11 @@
 package org.broadinstitute.listener.relay.http;
 
 import com.microsoft.azure.relay.RelayedHttpListenerContext;
+import java.io.ByteArrayInputStream;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.net.http.HttpResponse;
+import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
```diff
@@ -42,14 +44,27 @@ public OutputStream getCallerResponseOutputStream() {
...
     String statusDescription = "";
     if (ex != null && ex.getMessage() != null) {
       statusDescription = ex.getMessage();
     }
 
-    return new TargetHttpResponse(null, null, statusCode, statusDescription, context);
+    String body =
+        String.format(
+            "{ \"message\":\"The listener failed to process the request.\", \"tracking_id\":\"%s\"}",
+            context.getTrackingContext().getTrackingId());
```
Qi77Qi, line 57: shall we add the exception message to the response as well?

giventocode: I prefer not to return internal errors back to clients due to security concerns, also it may be meaningless to an end-user using the browser.  To help troubleshoot the issue a tracking id is provided. We can use it as a correlation id in the logs and see the details of what happened.

Qi77Qi: that sounds good....hopefully we'll be able to setup logging in a way that it'll be easy to access listener log

Qi77Qi (approved):

PR #14, opened 2022-05-24
IA-3448 add CORS headers

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3448

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
@@ -1,0 +1,3 @@
+package org.broadinstitute.listener.config;
+
+public record CorsSupportProperties(String preflightMethods) {}
--- a/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
@@ -1,3 +1,3 @@
 package org.broadinstitute.listener.config;
 
-public record CorsSupportProperties(String preflightMethods) {}
+public record CorsSupportProperties(String preflightMethods, String allowHeaders, String maxAge) {}
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
@@ -60,7 +60,8 @@ public class RelayedHttpRequestProcessor {
 
       clientResponse = httpClient.send(localRequest, HttpResponse.BodyHandlers.ofInputStream());
 
-      return TargetHttpResponse.createTargetHttpResponse(clientResponse, request.getContext());
+      return TargetHttpResponse.createTargetHttpResponse(
+          clientResponse, request.getContext(), corsSupportProperties);
 
     } catch (Throwable ex) {
 
@@ -109,18 +110,17 @@ public class RelayedHttpRequestProcessor {
         .getHeaders()
         .put(ACCESS_CONTROL_ALLOW_METHODS, corsSupportProperties.preflightMethods());
 
-    if (!listenerResponse.getHeaders().containsKey(ACCESS_CONTROL_ALLOW_ORIGIN))
-      listenerResponse
-          .getHeaders()
-          .put(
-              ACCESS_CONTROL_ALLOW_ORIGIN,
-              context.getRequest().getHeaders().getOrDefault("Origin", "*"));
+    listenerResponse
+        .getHeaders()
+        .put(
+            ACCESS_CONTROL_ALLOW_ORIGIN,
+            context.getRequest().getHeaders().getOrDefault("Origin", "*"));
 
     listenerResponse.getHeaders().put(ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
     listenerResponse
         .getHeaders()
-        .put(ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept, Origin,X-App-Id");
-    listenerResponse.getHeaders().put(ACCESS_CONTROL_MAX_AGE, "1728000");
+        .put(ACCESS_CONTROL_ALLOW_HEADERS, corsSupportProperties.allowHeaders());
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_MAX_AGE, corsSupportProperties.maxAge());
     try {
       listenerResponse.getOutputStream().close();
     } catch (IOException e) {
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
@@ -1,5 +1,11 @@
 package org.broadinstitute.listener.relay.http;
 
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS;
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS;
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_METHODS;
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN;
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_MAX_AGE;
+
 import com.microsoft.azure.relay.RelayedHttpListenerContext;
 import com.microsoft.azure.relay.RelayedHttpListenerResponse;
 import java.io.IOException;
@@ -18,7 +24,6 @@ import org.broadinstitute.listener.relay.transport.TargetResolver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.lang.NonNull;
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_METHODS;
 
 public class RelayedHttpRequestProcessor {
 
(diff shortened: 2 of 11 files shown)


service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
```diff
@@ -81,8 +86,15 @@ public static TargetHttpResponse createTargetHttpResponse(
...
                 String headerValue = value.iterator().next();
-                responseHeaders.put(key, headerValue);
+                if (headerValue != ACCESS_CONTROL_ALLOW_ORIGIN && headerValue != CONTENT_SECURITY_POLICY)
+                  responseHeaders.put(key, headerValue);
               });
+
+      responseHeaders.put(ACCESS_CONTROL_ALLOW_ORIGIN, "*");
+      responseHeaders.put(ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+      responseHeaders.put(ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept, Origin, X-App-Id");
+      responseHeaders.put(ACCESS_CONTROL_MAX_AGE, "1728000");
```
jdcanas, line 96: I believe we will need
`Access-Control-Allow-Methods` with `OPTIONS, GET,`. Leo sets all `OPTIONS, POST, PUT, GET, DELETE, HEAD, PATCH`, so might be safer to include all.

Not sure if we need to set the `X-Frame-Options` header as well...

Qi77Qi: yea, that's for preflight....which I didn't know where that'll be..but just figured it out with Jesus

service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
```diff
@@ -0,0 +1,3 @@
+package org.broadinstitute.listener.config;
+
+public record CorsSupportProperties(String preflightMethods) {}
```
giventocode, line 3: We should update the README with documentation for this property - its purpose and default value.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+      return Result.FAILURE;
+    }
+
+    RelayedHttpListenerResponse listenerResponse = context.getResponse();
+    listenerResponse.setStatusCode(204);
+    listenerResponse
+        .getHeaders()
+        .put(ACCESS_CONTROL_ALLOW_METHODS, corsSupportProperties.preflightMethods());
+
+    if (!listenerResponse.getHeaders().containsKey(ACCESS_CONTROL_ALLOW_ORIGIN))
```
giventocode, line 112: I don't think this check has any meaning/purpose. We are crafting the response, so the header won't be set in any case.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+      listenerResponse
+          .getHeaders()
+          .put(
+              ACCESS_CONTROL_ALLOW_ORIGIN,
+              context.getRequest().getHeaders().getOrDefault("Origin", "*"));
```
giventocode, line 117: As per the CORS spec, we should return a 400 status (Bad request), instead of  `*`, with `Reason: CORS header 'Origin' cannot be added to `when the Origin header missing`. Ref: [CORS header 'Origin'...](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSOriginHeaderNotAdded)

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
     return Result.FAILURE;
   }
 
+  public Result writePreflightResponse(RelayedHttpListenerContext context) {
+    if (context.getResponse() == null) {
+      logger.error("The context did not have a valid response");
+      return Result.FAILURE;
+    }
+
+    RelayedHttpListenerResponse listenerResponse = context.getResponse();
```
giventocode, line 106: maybe is another ticket, but we need to validate that the CORS request is valid - contains the expected headers and the requested URL is checked against a allowed list/setting - before creating a valid response.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+          .getHeaders()
+          .put(
+              ACCESS_CONTROL_ALLOW_ORIGIN,
+              context.getRequest().getHeaders().getOrDefault("Origin", "*"));
+
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+    listenerResponse
+        .getHeaders()
+        .put(ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept, Origin,X-App-Id");
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_MAX_AGE, "1728000");
```
giventocode, line 123: We should make the allowed headers and age configuration properties - in another ticket.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+    try {
+      listenerResponse.getOutputStream().close();
+    } catch (IOException e) {
+      logger.error("Failed to close response body to the remote client.", e);
+    }
+    return Result.FAILURE;
```
giventocode, line 129: It should return success instead of failure, if this line is reached.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
+    listenerResponse
+        .getHeaders()
+        .put(ACCESS_CONTROL_ALLOW_HEADERS, "Authorization, Content-Type, Accept, Origin,X-App-Id");
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_MAX_AGE, "1728000");
+    try {
+      listenerResponse.getOutputStream().close();
+    } catch (IOException e) {
+      logger.error("Failed to close response body to the remote client.", e);
+    }
```
giventocode, line 128: should return Failure if an exception occurred.

service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
```diff
@@ -83,6 +87,15 @@ public static TargetHttpResponse createTargetHttpResponse(
                 String headerValue = value.iterator().next();
                 responseHeaders.put(key, headerValue);
               });
+
```
giventocode, line 90: We need to remove this code. Otherwise, all responses will have these headers.

giventocode (changes requested):

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
Qi77Qi: if you're certain `context.getResponse()` will have empty header, I'm happy to remove the check

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
Qi77Qi: that's probably true if we're trying to enforce CORS..but here, we're basically trying to allow the restriction, let's say user is using a browser that doesn't send the header, are we just gonna say "Terra doesn't work on your browser"?

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
Qi77Qi: I think that's basically the same as `Referer` header check... @dinvlad is it necessary to check both `Referer` header and `Origin` header?

dinvlad: I would test it - I believe all browsers but IE provide `Referer`, but we may want to check `Origin` also if `Referer` is not available (but `Origin` may not be provided, based on the context, so please test it in the field)

service/src/main/java/org/broadinstitute/listener/relay/http/TargetHttpResponse.java
Qi77Qi: I thikn we do need these....otherwise things don't work

Qi77Qi: I'll try if removing works

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
giventocode: A few more details/thoughts on this: `Origin` is a required header as per the CORS algorithm spec details [here](https://www.w3.org/TR/2020/SPSD-cors-20200602/#resource-preflight-requests).  So if it is missing, or not valid, the request must be terminated.  A `Referer` check is outside the scope of CORS from what I can tell.  If we have a functional need to check the `Referer`, IMO it is best to implement the check as an `inspector` and not part of the CORS validation process.

dinvlad: I think whichever option works across browsers, it doesn't matter that much. Please use this as a reference wrt CSRF protection: https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#verifying-origin-with-standard-headers

dinvlad: Also, if Origin works then it might be slightly _preferred_, because without a proper Referrer-Policy header, Referer might contain URL path in addition to the origin itself, and then the path needs to be parsed out, which adds room for error. Better keep it simple with Origin, if possible.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -83,6 +97,38 @@ public Result writeNotAcceptedResponseOnCaller(RelayedHttpListenerContext contex
...
+        .getHeaders()
+        .put(ACCESS_CONTROL_ALLOW_METHODS, corsSupportProperties.preflightMethods());
+
+    listenerResponse
+        .getHeaders()
+        .put(
+            ACCESS_CONTROL_ALLOW_ORIGIN,
+            context.getRequest().getHeaders().getOrDefault("Origin", "*"));
+
+    listenerResponse.getHeaders().put(ACCESS_CONTROL_ALLOW_CREDENTIALS, "true");
```
dinvlad, line 118: Just one more comment on this - I think technically these 2 headers are not allowed by browsers to be mixed, i.e. even if the server responds with both, credentials will not be sent when server allows `*`. So allowing credentials here is kinda useless :)

PR #15, opened 2022-06-02
IA-3416 Add setCookie API in listener

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3416


service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -0,0 +1,25 @@
...
+
+  public static final Optional<String> getTokenFromAuthorization(Map<String, String> headers) {
+    var authValue = headers.getOrDefault(AUTHORIZATION, null);
+
+    return Optional.ofNullable(authValue)
+        .filter(s -> s.contains("Bearer "))
+        .map(s -> s.replaceFirst("Bearer ", "").trim());
+  }
+
+  public static boolean isSetCookiePath(java.net.URI uri) {
```
giventocode, line 20: nit: full type, an import statement is better.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -0,0 +1,25 @@
...
+    var splitted = uri.getPath().split("/");
+    if (splitted.length == 3) return splitted[2].equals(SET_COOKIE_API_PATH);
```
giventocode, line 22: Is there a case where you would get http://foo.com, instead of http://foo.com/? If so, the implementation won't work.

service/src/main/java/org/broadinstitute/listener/relay/transport/RelayedRequestPipeline.java
```diff
@@ -78,6 +78,8 @@ public void registerHttpExecutionPipeline() {
         .receiveRelayedHttpRequests()
         .filter(c -> listenerConnectionHandler.isNotPreflight(c.getRequest()))
         .doOnDiscard(RelayedHttpListenerContext.class, httpRequestProcessor::writePreflightResponse)
+        .filter(c -> listenerConnectionHandler.isNotSetCookie(c.getRequest()))
```
giventocode, line 81: I thought we wanted this to be executed after the inspectors.

giventocode (changes requested):

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
Qi77Qi: it'll work...
```
@ java.net.URI.create("https://relay-ns-995014ac-a05d-477e-9a75-1aa10bd5def5.servicebus.windows.net/saturn-4c90d242-dc9d-4d26-b858-4b7c73a97efc/setCookie/").getPath().split("/")
res1: Array[String] = Array("", "saturn-4c90d242-dc9d-4d26-b858-4b7c73a97efc", "setCookie")

vs

@ java.net.URI.create("https://relay-ns-995014ac-a05d-477e-9a75-1aa10bd5def5.servicebus.windows.net/saturn-4c90d242-dc9d-4d26-b858-4b7c73a97efc/setCookie").getPath().split("/")
res3: Array[String] = Array("", "saturn-4c90d242-dc9d-4d26-b858-4b7c73a97efc", "setCookie")
```

But I'll add a unit test as well

giventocode: Okay, I misunderstood the method, what is happening here is a check if we are dealing with a set cookie request.  Adding the unit tests would be good.

giventocode (approved):

PR #16, opened 2022-06-02
[TOAZ-157] Fix: Removes transfer-encoding header, that was causing conn-resets with Azure Relay

giventocode:


service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -176,6 +179,11 @@ public Result writeTargetResponseOnCaller(@NonNull TargetHttpResponse targetResp
     return result;
   }
 
+  private void removeHeadersNotAcceptedByAzureRelay(Map<String, String> headers) {
+    headers.remove("transfer-encoding");
```
Qi77Qi, line 183: is there a way to make this case insensitive?

giventocode: hmm, I tried to cover the two most typical cases, it is very unlikely that we get all caps or other strange combination. To make it case insensitive, I would have to iterate through all the keys in the map and do a case insensitive comparison for each key - seems like overkill for this...

Qi77Qi (approved):
once tests pass

PR #17, opened 2022-06-03
BT-654 Make Sam resource type configurable

jgainerdewar:
I tested this by deploying it to our Cromwell app using the changes in https://github.com/broadinstitute/cromwhelm/pull/31 and configuring it to use workspace access when checking permission.

Unsure whether retaining the old sam resource type value as a default is valuable, I just didn't want Leo's use of this to break.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ that have write access to a specific Sam resource.
 `listener.samInspectorProperties.samResourceId`: The id of the Sam resource to check access to
 
 `listener.samInspectorProperties.samResourceType`: The type of the Sam resource to check access to.
-Defaults to `controlled-application-private-workspace-resource`, which corresponds to Jupyter Notebooks.
+Defaults to `controlled-application-private-workspace-resource`, which corresponds Leo-managed resources.
 
 ## Running Jupyter Notebooks
 
@@ -44,7 +44,7 @@ the following configuration settings are required.
 
 `c.NotebookApp.allow_origin = '*'`
 
-*Note*: You could add an specific Azure Relay origin if required.
+*Note*: You could add a specific Azure Relay origin if required.
 
 `c.NotebookApp.base_url = '/<HYBRID CONNECTION NAME>/'`
 
--- a/README.md
+++ b/README.md
@@ -25,6 +25,18 @@ The Terra Azure Relay Listener establishes a bi-directional channel with Azure R
 
 `listener.corsSupportProperties.preflightMethods` Methods that we support CORS. Default to `OPTIONS, POST, PUT, GET, DELETE, HEAD, PATCH`.
 
+### Sam Inspector config options
+
+By using the Sam Checker inspector, the Listener can be configured to allow access only for users
+that have write access to a specific Sam resource.
+
+`listener.samInspectorProperties.samUrl`: URL to the Sam instance we should talk to
+
+`listener.samInspectorProperties.samResourceId`: The id of the Sam resource to check access to
+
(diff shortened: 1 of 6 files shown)


service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
```diff
@@ -1,3 +1,31 @@
 package org.broadinstitute.listener.config;
 
-public record SamInspectorProperties(String samUrl, String samResourceId) {}
```
Qi77Qi, line 3: why not just use `record`?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -34,7 +38,7 @@ public Instant checkWritePermission(String accessToken) {
         samClient.setAccessToken(accessToken);
         var resourceApi = new ResourcesApi(samClient);
 
-        var res = resourceApi.resourcePermissionV2(SAM_RESOURCE_TYPE, samResourceId, "write");
```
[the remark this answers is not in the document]
Qi77Qi: or is this just to make it easier to test things outside leonardo?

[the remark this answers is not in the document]
jgainerdewar: Eventually it will be `kubernetes-app` (IIRC). Right now we're using `workspace`.

service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
jgainerdewar: I would have preferred that, but it looks like I can't get the default `samResourceId` set in here if it's a record. Happy to hear better ideas, I'm new to modern Java and may be missing obvious ways to do things.

Qi77Qi: maybe becuz https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/main/service/src/main/resources/application.yml#L18 is not updated?

Btw, I think should update this yml file with the new key...and having the leo specific key as default. We've also been documenting configs in https://github.com/DataBiosphere/terra-azure-relay-listeners#configuration-properties

jgainerdewar: Ah, I got distracted by the nifty environment variables and forgot to update the actual config file. I'll set the default there and revert this to `record`. Will also add information about the sam inspector to the README.

README.md
```diff
@@ -25,6 +25,18 @@ The Terra Azure Relay Listener establishes a bi-directional channel with Azure R
...
+
+By using the Sam Checker inspector, the Listener can be configured to allow access only for users
+that have write access to a specific Sam resource.
+
+`listener.samInspectorProperties.samUrl`: URL to the Sam instance we should talk to
+
+`listener.samInspectorProperties.samResourceId`: The id of the Sam resource to check access to
+
+`listener.samInspectorProperties.samResourceType`: The type of the Sam resource to check access to.
+Defaults to `controlled-application-private-workspace-resource`, which corresponds to Jupyter Notebooks.
```
Qi77Qi, line 38: I think this corresponds to leo managed resources...for cromwell, it'll be the same when managed through leonardo

Qi77Qi (commented):
lgtm...were you able to test this out in some way?

Qi77Qi (approved):

breilly2 (approved):

PR #19, opened 2022-09-08
IA-3698 Implementation of routing rules

giventocode:
- The listener supports routing to targets based on configuration of rules.
 - A rule contains a `containsPath` String and the `targetHost`
 - Documentation update with samples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The Terra Azure Relay Listener establishes a bi-directional channel with Azure R
 
 `listener.corsSupportProperties.preflightMethods` Methods that we support CORS. Default to `OPTIONS, POST, PUT, GET, DELETE, HEAD, PATCH`.
 
-`listener.targetProperties.targetRoutingRules` As list of target routing rules. A rule is a tuple of the string to look in the URI and the target host.
+`listener.targetProperties.targetRoutingRules` A list of routing rules. A rule is a tuple of the string the URI must contain for a match and the target host.
 The default `targetHost` is used. Example configuration:
 
 ```yaml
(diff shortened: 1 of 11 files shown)


service/src/main/java/org/broadinstitute/listener/relay/transport/DefaultTargetResolver.java
```diff
@@ -59,6 +56,23 @@ public URL createTargetUrl(@NonNull URI relayedRequestUri) throws InvalidRelayTa
...
+    Optional<String> ruleTargetHost =
+        rules.stream()
+            .filter(r -> relayedRequestUri.toString().contains(r.pathContains()))
+            .map(r -> r.targetHost())
+            .findFirst();
+
+    if (ruleTargetHost.isPresent()) {
+      return ruleTargetHost.get();
+    }
+    return defaultTargetHost;
```
pshapiro4broad, line 73: In most cases you shouldn't need to call `isPresent()` followed by `get()`.  If you don't have any `else` behavior, use `map()`. If you have an `else` behavior, use one of the `orElse()` methods. In this case:
```suggestion
    return ruleTargetHost.orElse(defaultTargetHost);
```

README.md
```diff
@@ -17,14 +17,25 @@ The Terra Azure Relay Listener establishes a bi-directional channel with Azure R
...
+`listener.targetProperties.targetRoutingRules` A list of routing rules. A rule is a tuple of the string the URI must contain for a match and the target host.
+The default `targetHost` is used. Example configuration:
+
+```yaml
+  targetProperties:
+    targetHost: "http://localhost:8080"
+    targetRoutingRules:
+      -
+        pathContains: "welder"
+        targetHost: "http://localhsot:8081"
```
Qi77Qi, line 37: how will `listener.targetProperties.targetRoutingRules` look like in env? is it going to be sth like `LISTENER_TARGETPROPERTIES_TARGETROUTINGRULES_0_PATHCONTAINS = "welder"`? anyway we can make it sth that doesn't involve `_<number>`?

Qi77Qi (commented):
can we add the keys in application.yml as well

Qi77Qi (approved):

PR #22, opened 2022-11-03
[TOAZ-242]  Makes sam action configurable, and tests fixes.

giventocode:
- Added the `listener.samInspectorProperties.samAction` config property, with a default value of `write`
 - Fixed tests that were not passing prior to the change.

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
@@ -1,3 +1,42 @@
 package org.broadinstitute.listener.config;
 
-public record SamInspectorProperties(String samUrl, String samResourceId, String samResourceType) {}
+public final class SamInspectorProperties {
+
+  private static final String DEFAULT_SAM_ACTION = "write";
+  private String samUrl;
+  private String samResourceId;
+  private String samResourceType;
+  private String samAction = DEFAULT_SAM_ACTION;
+
+  public String getSamUrl() {
+    return samUrl;
+  }
+
+  public String getSamResourceId() {
+    return samResourceId;
+  }
+
+  public String getSamResourceType() {
+    return samResourceType;
+  }
+
+  public String getSamAction() {
+    return samAction;
+  }
+
+  public void setSamAction(String samAction) {
+    this.samAction = samAction;
+  }
+
+  public void setSamUrl(String samUrl) {
+    this.samUrl = samUrl;
+  }
+
+  public void setSamResourceId(String samResourceId) {
+    this.samResourceId = samResourceId;
+  }
+
+  public void setSamResourceType(String samResourceType) {
+    this.samResourceType = samResourceType;
+  }
+}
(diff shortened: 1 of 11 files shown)


service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
```diff
@@ -1,3 +1,42 @@
 package org.broadinstitute.listener.config;
 
-public record SamInspectorProperties(String samUrl, String samResourceId, String samResourceType) {}
+public final class SamInspectorProperties {
```
Qi77Qi, line 3: any reason why converting this into a class instead of having it as a `record`? if it's just for default value, I think we can set it in application.yml?

Qi77Qi (commented):
would be nice if you could test it out in a BEE...if not, one of us should test it out before merging ideally

service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
giventocode: As a record it'd require the property to be set in the application.yml explicitly.  I wanted the value to be truly optional, that is, not required at all in the application.yml so existing config works with this version. However, if this is not a concern, and we are certain all services that use this version will update their config, I can revert it to a record, make the property required, and add it to the application.yml file.

rtitle: I think all the listener configuration comes in the form of environment variables, e.g. setting `LISTENER_SAMINSPECTORPROPERTIES_SAMRESOURCEID=...` at listener deployment time. In other words, there aren't any clients specifying their own `application.yml`. So it seems safe to put `samAction: write` in application.yml (we can test).

rtitle (approved):
Looks good! I can help test this out from the Leo side (probably tomorrow).

service/src/main/java/org/broadinstitute/listener/config/SamInspectorProperties.java
Qi77Qi: yea, I like having all configs in application.yml just so I dont have to also look at other places about optional configs

Qi77Qi (approved):
if things look good after testing 👍

rtitle:
PRs to use this:
https://github.com/broadinstitute/cromwhelm/pull/110
https://github.com/DataBiosphere/leonardo/pull/2979

Will test it out next week.

rtitle:
It works! On the relaylistener:
```
root@relaylistener-8465d8f64d-wj2sq:/# env | grep SAM
LISTENER_REQUESTINSPECTORS=SAM_CHECKER
LISTENER_SAMINSPECTORPROPERTIES_SAMACTION=connect
LISTENER_SAMINSPECTORPROPERTIES_SAMRESOURCETYPE=kubernetes-app
LISTENER_SAMINSPECTORPROPERTIES_SAMURL=https://sam.bee-rtitle-busy-kangaroo.bee.envs-terra.bio/
LISTENER_SAMINSPECTORPROPERTIES_SAMRESOURCEID=dc476cd1-e714-4ea7-b31a-53b978ab5466
```

And I confirmed that I can access that resource from Leo.

PR #23, opened 2022-11-04
IA-3822 don't permission check on status calls

Qi77Qi:
https://broadworkbench.atlassian.net/browse/IA-3822

Can't add unit test easily becuz `RelayedHttpListenerRequest` is package priavte in azure SDK

Leo needs to make `status` calls to check if app or welder etc is up & running. Right now, these calls will be blocked by permission check, but we really don't need to check permission for `status` calls (maybe we need a different pattern for apps since not all apps have a `status` api...but this PR at least take care of `status`)

Will test things out

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamPermissionInspector.java
@@ -30,7 +30,9 @@ public class SamPermissionInspector implements RequestInspector {
   @Override
   public boolean inspectRelayedHttpRequest(
       @NonNull RelayedHttpListenerRequest relayedHttpListenerRequest) {
-    return checkPermission(relayedHttpListenerRequest.getHeaders());
+    var requestPath = relayedHttpListenerRequest.getUri().toString();
+    if (requestPath.endsWith("status") || requestPath.endsWith("status/")) return true;
+    else return checkPermission(relayedHttpListenerRequest.getHeaders());
   }
 
   private boolean checkPermission(Map<String, String> headers) {


rtitle (commented):
I don't think we should make it possible to make unauthenticated calls to an app. Right now there's a hole -- imagine there's a 3rd party app and they define a `/something/status` endpoint, it would bypass our security.

Can we pass the user token on status checks to welder/jupyter? That's what we're trying to do for Azure apps -- [this PR](https://github.com/DataBiosphere/leonardo/pull/2925) has a solution using a pet token.

Qi77Qi:
what's the downside of exposing `status` api tho? r u thinking some apps might container sensitive data in their `status` api?

if that's the case, I can also also limit it to just `welder`...welder's status api just has the version

rtitle:
I'm more worried about poking holes. Every app needs status checking right? Welder, Jupyter, RStudio, Cromwell, custom apps all need to poll status. Do we need to define an unauthenticated `/status` endpoint for each of those? What about 3rd party apps where we don't control the code? It adds some level of trust to the apps.

I think it's a better security model if all https traffic to Leo-apps is controlled by Sam. What's the downside of just passing a pet token from Leo?

Qi77Qi:
mostly I'm just comparing to how we do `status` endpoint in control plane services like Leonardo, Sam etc today...all the status endpoints are auth checked..

If you feel strongly about this, I can look into passing the token

rtitle:
Yeah thanks, I think apps/runtimes are a little different than control plane services in this regard…

Just looking at code, looks like [HttpJupyterDAO](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpJupyterDAO.scala#L27) passes the Leo token for the Azure runtimes, which probably works based on how the Sam resource is setup. So presumably [HttpWelderDAO](https://github.com/DataBiosphere/leonardo/blob/develop/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/dao/HttpWelderDAO.scala) could do the same?

Qi77Qi:
yea, I'm doing that

PR #25, opened 2023-01-25
[AJ-805] File uploads block subsequent requests when using Jupyter labs 

giventocode:
- The cause of the issue was a blocking condition encountered when the SDK flushes/closes the connection after a PUT operation. 
 - This PR configures the reactive pipeline not to use the default thread pool.


rtitle (approved):
Looks reasonable. 👍 

I can try to quickly test this in my environment

rtitle:
Tested in my notebook environment, works great

PR #26, opened 2023-02-28
AJ-869 fix blocking calls

calypsomatic:
Long-lasting requests through the relay listener were blocking subsequent requests to the same service.   Seems the listener was not using the scheduler as expected; this PR forces it to do so.


jordansafer (approved):

giventocode (approved):
LGTM

davidangb:
fwiw I am also worried about the call to Sam, since we use the Sam client in a blocking fashion - and I think your local testing bypassed Sam. This PR is great as-is and is a definite improvement; we might ALSO want to assess/fix the Sam call in a separate PR.

PR #27, opened 2023-03-20
[IA-4125] Set accessed date via inspector

giventocode:
This PR includes the following:
- New listener inspector `SetAccessedDateInspector` that calls Leo using the user auth token when a request is received within a window.
- At the most 1 request to Leo is made within a given window if more requests are received.
- The inspector won't block the request if the call to Leo fails. 
- Refactoring, inspectors are now executed in the same thread as the request to the local endpoint and not on the main thread. 
- Updated documentation with configuration values for the new inspector.

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
@@ -94,12 +94,17 @@ public class SetDateAccessedInspector implements RequestInspector {
     return checkLastAccessDateAndCallServiceIfExpired(relayedHttpListenerRequest);
   }
 
-  private synchronized boolean checkLastAccessDateAndCallServiceIfExpired(
+  private boolean checkLastAccessDateAndCallServiceIfExpired(
       RelayedHttpListenerRequest relayedHttpListenerRequest) {
-    if (hasLastAccessDateExpired()) {
-      setLastAccessedDateOnService(relayedHttpListenerRequest);
-      updateLastAccessedDate();
+    try {
+      if (hasLastAccessDateExpired()) {
+        setLastAccessedDateOnService(relayedHttpListenerRequest);
+        updateLastAccessedDate();
+      }
+    } catch (RuntimeException ex) {
+      logger.error("Failed set the last accessed date. The request still will get processed", ex);
     }
+
     return true;
   }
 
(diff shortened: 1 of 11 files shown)


LizBaldo (approved):
LGTM! Thanks a bunch for the updated documentation, I can follow the logic, I only have a small question about `inspectWebSocketUpgradeRequest` vs, `inspectRelayedHttpRequest` as they seem to be doing the exact same thing 🤔

giventocode:
> LGTM! Thanks a bunch for the updated documentation, I can follow the logic, I only have a small question about `inspectWebSocketUpgradeRequest` vs, `inspectRelayedHttpRequest` as they seem to be doing the exact same thing 🤔

Great! The listener supports websocket and http connections. These scenarios result in different callbacks from the SDK and trigger different pipelines in the listener’s implementation. The first step in a websocket connection is an http upgrade request. The inspectWebSocketUpgradeRequest method is called only when a websocket upgrade request is received. This enables an inspector to accept or block the subsequent websocket connection. Having a separate method for the upgrade request gives the flexibility in case you want to handle/inspect these requests differently than regular http requests. For this inspector, however, it makes sense to handle websocket upgrade requests as any other, hence having the same code.

A bit more info on the inspectors [here](https://github.com/DataBiosphere/terra-azure-relay-listeners#request-inspectors)

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Locale;
+import org.apache.http.client.utils.URIBuilder;
+import org.broadinstitute.listener.relay.Utils;
+import org.broadinstitute.listener.relay.inspectors.InspectorType.InspectorNameConstants;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component(InspectorNameConstants.SET_DATE_ACCESSED)
+public class SetDateAccessedInspector implements RequestInspector {
```
rtitle, line 22: Quick question @giventocode, does `SetDateAccessedInspector` _always_ run in the listener, meaning the new config options are required?

Currently the listener is deployed in front of runtimes and apps; I think at the moment we only want to call `setDateAccessed` for runtime deployments since that is where we have [auto-]pause implemented. We probably don't want apps to be calling `/api/v2/runtimes/%s/%s/setDateAccessed` (in the future we may want to auto-pause apps as well).

Would it be possible for Leo to enable/disable `SetDateAccessedInspector` as needed at deployment time?

giventocode: Yeah, good question. The inspectors can be turn on/off via config. I updated the properties with the new properties for the `setDateAccessed` inspector, but with the inspector **disabled** .

https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/main/service/src/main/resources/application.yml

```yaml
  requestInspectors:
#    - headersLogger
    - samChecker
#    - setDateAccessed
```

rtitle: Oh gotcha.. missed that, thank you! 

(Maybe at some point we can refactor this to use events instead of direct listener->Leo coupling ;))

PR #29, opened 2023-04-28
[IA-4082] Update access control allow origin and cookie.

n8landolt:
__Jira ticket__: https://broadworkbench.atlassian.net/browse/IA-4081

Leo PR:
https://github.com/DataBiosphere/leonardo/pull/3311

### What

- Update reference.conf to point to the image generated from the TARL PR

### Why

- Security vulnerability with request headers.

### What to test

- Inspect headers on a request with a Cookie, and confirm it ends with 'HttpOnly'
- Inspect headers on a relayed HTTP request and confirm AccessControlAllowOrigin is '*'

How to test:
Check out on your bee and deploy the Leo image for the branch `IA-4081-update-azure-relay-listeners` version hash `74fb80a9a19b` using the terra helmfile version: `0.169.101`. Details of this helmfile version can be found here: https://github.com/broadinstitute/terra-helmfile/pull/4184 
The helmfile changes remove the wildcard from the allowed hosts, which effectively ignores my Relay Listener Changes. I also included the BEE Urls for the reviewers.

Copy a CURL request to the relay listener endpoint (Such as storageLinks) and edit it to have a "bogus" origin.

The Server should deny the request.

### Who tested and where

- [ ] This change is covered by automated tests 
  - _NB: Rerun automation tests on this PR by commenting `jenkins retest` or `jenkins multi-test`._
- [ ] I validated this change 
- [ ] Primary reviewer validated this change 
- [ ] I validated this change in the dev environment

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
@@ -1,4 +1,10 @@
 package org.broadinstitute.listener.config;
 
+import java.util.List;
+
 public record CorsSupportProperties(
-    String preflightMethods, String allowHeaders, String maxAge, String contentSecurityPolicy) {}
+    String preflightMethods,
+    String allowHeaders,
+    String maxAge,
+    String contentSecurityPolicy,
+    List<String> validHosts) {}
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -1,20 +1,21 @@
 package org.broadinstitute.listener.relay;
 
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_CREDENTIALS;
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_HEADERS;
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_METHODS;
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN;
-import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_MAX_AGE;
-import static com.google.common.net.HttpHeaders.AUTHORIZATION;
-import static com.google.common.net.HttpHeaders.CONTENT_SECURITY_POLICY;
-
+import org.broadinstitute.listener.config.CorsSupportProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.net.MalformedURLException;
 import java.net.URI;
+import java.net.URL;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
-import org.broadinstitute.listener.config.CorsSupportProperties;
+import java.util.regex.Pattern;
+
+import static com.google.common.net.HttpHeaders.*;
 
 public class Utils {
+
+  public static final Logger logger = LoggerFactory.getLogger(Utils.class);
   public static final String TOKEN_NAME = "LeoToken";
   public static final String SET_COOKIE_API_PATH = "setcookie";
 
@@ -28,8 +29,11 @@ public class Utils {
 
   public static boolean isSetCookiePath(URI uri) {
     var splitted = uri.getPath().split("/");
-    if (splitted.length == 3) return splitted[2].toLowerCase().equals(SET_COOKIE_API_PATH);
-    else return false;
+    if (splitted.length == 3) {
+      return splitted[2].toLowerCase().equals(SET_COOKIE_API_PATH);
+    } else {
+      return false;
+    }
   }
 
   public static void writeCORSHeaders(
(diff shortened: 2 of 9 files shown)


LizBaldo (commented):
LGTM, let me know when you want a 👍 so it is ready to test

mlilynolting (commented):
lgtm2

service/src/main/resources/application.yml
```diff
@@ -37,3 +37,4 @@ listener:
     allowHeaders: "Authorization, Content-Type, Accept, Origin,X-App-Id"
     maxAge: "1728000"
     contentSecurityPolicy: "frame-ancestors http://localhost:3000;report-uri https://terra.report-uri.com/r/d/csp/reportOnly"
+    validHosts:
```
jdcanas, line 40: what does the empty inclusion here do, is this populated elsewhere?

service/src/test/java/org/broadinstitute/listener/relay/http/TargetHttpResponseTest.java
```diff
@@ -74,9 +81,18 @@ void createLocalHttpResponse_containsValidResponse() {
...
+      List<String> validHosts =
+          new ArrayList<>() {
+            {
+              add("*");
+            }
+          };
+      targetHttpResponse =
+          targetHttpResponse.createTargetHttpResponse(
+              httpResponse, context, new CorsSupportProperties("", "", " ", "", validHosts));
+    } catch (Throwable ex) {}
```
jdcanas, line 94: should we explicitly fail in the catch block?

service/src/main/resources/application.yml
n8landolt: Not entirely sure - I had seen it elsewhere (Can't remember off the top of my head) but I figured it's a reminder placeholder that we expect a value there.

Do you have a suggestion for a default value?

service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
```diff
@@ -1,4 +1,10 @@
...
 
+import java.util.List;
+
 public record CorsSupportProperties(
-    String preflightMethods, String allowHeaders, String maxAge, String contentSecurityPolicy) {}
+    String preflightMethods,
+    String allowHeaders,
+    String maxAge,
+    String contentSecurityPolicy,
+    List<String> validHosts) {}
```
LizBaldo, line 10: Where is the allowlist of valid hosts defined? Is it in the yaml file below?

service/src/main/resources/application.yml
LizBaldo: Is this where we should define the valid hosts allowlist?

n8landolt: We COULD have some default values here, but it will be overwritten by arguments passed to the listener's docker command.

Personally, I think it's at least worthwhile to keep here as a reminder that it needs values.

service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
n8landolt: This is defined in Leonardo, and is actually added to the listener start command. We also add the SERVER_APP_WEBSOCKET_HOST which is generated some time during VM creation.

https://github.com/DataBiosphere/leonardo/pull/3311/files#diff-375ff112a6898504a9074d258d167ba63d6300124c1438b0351561eb7b8fc4ecR242

service/src/test/java/org/broadinstitute/listener/relay/UtilTest.java
```diff
@@ -49,4 +51,31 @@ void isSetCookiePath() {
     var res4 = Utils.isSetCookiePath(uri4);
     assertThat(res4, equalTo(true));
   }
+
+  @Test
+  void isValidOrigin_test() {
```
mlilynolting, line 56: I think a few more test cases would be best, especially if the regex matching is left in.

mlilynolting (commented):
Haven't run this yet but did have some feedback!

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -47,4 +50,28 @@ public static void writeCORSHeaders(
...
+      stringToCheck = origin;
+    }
+
+    // To appeal to lambda.
+    String finalStringToCheck = stringToCheck;
+    return origin.isEmpty()
+        || corsSupportProperties.validHosts().stream()
+            .anyMatch(
+                validHost ->
+                    Pattern.matches(
```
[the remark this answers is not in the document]
n8landolt: The alternate approach is checking if valid hosts has a wildcard, if so, pass, else, check contains?

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -47,4 +50,28 @@ public static void writeCORSHeaders(
     responseHeaders.putIfAbsent(ACCESS_CONTROL_ALLOW_HEADERS, corsSupportProperties.allowHeaders());
     responseHeaders.putIfAbsent(ACCESS_CONTROL_MAX_AGE, corsSupportProperties.maxAge());
   }
+
+  public static boolean isValidOrigin(String origin, CorsSupportProperties corsSupportProperties) {
+    String stringToCheck;
+    // We want to strip the protocol.
```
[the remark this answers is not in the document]
n8landolt: I'm going to go with since the ticket didn't specify schema, and the code base doesn't have an "allowed schemas" list, that we don't have to worry about it in this ticket.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -47,4 +50,28 @@ public static void writeCORSHeaders(
...
+    try {
+      URL url = new URL(origin);
+      stringToCheck = url.getHost();
```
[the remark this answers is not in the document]
n8landolt: Let me write a test for this scenario! Good catch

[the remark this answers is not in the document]
n8landolt: According to my test, the port is included in host

service/src/main/java/org/broadinstitute/listener/config/CorsSupportProperties.java
LizBaldo: Ah that makes a lot of sense, thanks!

LizBaldo (approved):
Works (and by that I mean I cannot post relay listener requests with a bad origin) for me!

PR #30, opened 2023-05-10
[IA-4308] Add status endpoint 

n8landolt:
so we can poll for listener ready.

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/StatusService.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/StatusService.java
@@ -1,0 +1,14 @@
+package org.broadinstitute.listener.relay.http;
+
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class StatusService {
+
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/broadinstitute/listener/relay/http/StatusService.java
```diff
@@ -0,0 +1,14 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class StatusService {
+
+  @GetMapping("/status")
+  public ResponseEntity<String> getStatus() {
+    return ResponseEntity.ok("Application is running.");
```
rtitle, line 12: Addmittedly we don't really have a spec for `/status`, but other services typically return a json structure:
```
{
  "ok": "true",
  "systems": {…}
}
```
Ex https://bpm.dsde-dev.broadinstitute.org/status

Not sure if listener has any subsystems it wants to include in its status check (guessing probably not).

n8landolt: This PR actually isn't necessary anymore after I did some digging. Forgot to clean it up!

PR #31, opened 2023-05-18
[WOR-1030] Update request logging

aherbst-broad:
We need to log additional information as part of relay requests for better auditability. This PR adds those elements and does a bit of cleanup + testing.


service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
```diff
@@ -20,27 +24,68 @@ public class HeaderLoggerInspector implements RequestInspector {
...
+    String endpoint = "";
+    if (relayedHttpListenerRequest.getRemoteEndPoint() != null) {
+      endpoint = relayedHttpListenerRequest.getRemoteEndPoint().toString();
+    }
 
-    logger.debug("HTTP Request: {}", relayedHttpListenerRequest.getUri());
+    var sub = getTokenClaim(relayedHttpListenerRequest.getHeaders(), "sub").orElse("");
+    var idTyp = getTokenClaim(relayedHttpListenerRequest.getHeaders(), "idtyp").orElse("");
+
+    logger.info(
```
aherbst-broad, line 56: This is an attempt to get the request logged into something apache-ish.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -47,4 +48,20 @@ public static void writeCORSHeaders(
     responseHeaders.putIfAbsent(ACCESS_CONTROL_ALLOW_HEADERS, corsSupportProperties.allowHeaders());
     responseHeaders.putIfAbsent(ACCESS_CONTROL_MAX_AGE, corsSupportProperties.maxAge());
   }
+
+  public static Optional<String> getToken(Map<String, String> headers) {
+    return Utils.getTokenFromCookie(headers).or(() -> Utils.getTokenFromAuthorization(headers));
+  }
+
+  private static Optional<String> getTokenFromCookie(Map<String, String> headers) {
```
rtitle, line 56: I know this is moved code -- just wondering if we should use `RelayedHttpRequest.getCookies()` instead of trying to parse the cookie value from the headers. Wondering if there are checks for expiration, etc that we're missing here..

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
rtitle: Maybe a comment would be useful that we're trying to mirror the apache format?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
```diff
@@ -20,27 +24,69 @@ public class HeaderLoggerInspector implements RequestInspector {
...
+    return true;
+  }
+
+  @VisibleForTesting
+  public void logRequest(
+      RelayedHttpListenerRequest relayedHttpListenerRequest,
+      OffsetDateTime requestTimestamp,
+      String prefix) {
+    var headers = relayedHttpListenerRequest.getHeaders();
+    var referer = headers.getOrDefault("Referer", "");
```
rtitle, line 44: Are there any case sensitivity issues here?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
```diff
@@ -20,27 +24,69 @@ public class HeaderLoggerInspector implements RequestInspector {
...
+    var origin = headers.getOrDefault("Origin", "");
+    var ua = headers.getOrDefault("User-Agent", "");
+    var sub = getTokenClaim(relayedHttpListenerRequest.getHeaders(), "sub").orElse("");
+
+    // get the idtyp so we can log if the request was made by a uami (i.e,. the value is "app")
+    var idTyp = getTokenClaim(relayedHttpListenerRequest.getHeaders(), "idtyp").orElse("");
```
rtitle, line 50: This claim probably works.. I think in apache we use the `xms_mirid` claim to determine if it's a UAMI (the claim contains the UAMI objectId).

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -41,12 +41,27 @@ public OauthInfo getOauthInfo(String token) throws IOException, InterruptedExcep
     }
   }
 
+  public Optional<String> getClaim(String token, String claim) {
+    try {
+      var decoded = JWT.decode(token);
```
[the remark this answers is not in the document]
rtitle: Er, I guess google tokens may not have all the claims we need but it should have `sub` probably..

rtitle (commented):
Few comments/questions above -- overall looks good though, thanks for cleaning up some of the logging/testing too.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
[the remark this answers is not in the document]
dinvlad: Hmm can a Google token be used against Relay? Or I guess we should be detecting when it is, as another an unexpected scenario (and literally "Use of service accounts for interactive log on" in the compliance checklist)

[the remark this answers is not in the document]
dinvlad: And more specifically, we could detect that based on `.gserviceaccount.com` suffix in the `email` claim.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
dinvlad: Hmm not sure I see status code here - we'd need it to detect "Use of service accounts for interactive log on"

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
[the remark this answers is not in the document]
rtitle: I mean technically any valid Terra token could be used against Relay. Possibilities:

1. Google user token -- e.g. if someone does `gcloud auth print-access-token` and makes a request to Relay. This is probably not common, I can't think of a Terra use case that would result in this.
2. Google service account token -- e.g. Terra control plane service calling relay. This happens during Leo polling, potentially other places.
3. B2C user token -- user opening an app in Terra UI
4. UAMI token -- data plane apps talking to each other, example CBAS and WDS

aherbst-broad:
@davidangb yeah great idea; @dinvlad mentioned needing something similar, I'll look into it.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
aherbst-broad: Yeah, good catch, will fix. Annoyingly the headers map is <string, string> so I'll convert to a treemap and a case insensitive compare that way.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
aherbst-broad: Yeah, the need for a post response chain complicates things. I'll look into it.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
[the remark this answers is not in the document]
aherbst-broad: Good catch @rtitle , I'll look into reusing the decoded + google token fallback we're using the sam stuff. So, it sounds like we'll want to emit the email claim (if present), will that suffice for the pet detection @dinvlad ?

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
rtitle: I thought I saw it googling but now opened Intellij and don't see it either in `com.microsoft.azure.relay` classes. The current approach is probably the right thing to do then.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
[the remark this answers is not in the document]
dinvlad: I think it'd be fine to log the email tbh, and we can always improve that later if there's a privacy concern. It'd be easier to review the alerts this way, although it's not strictly required for non-control plane SA detection (since we could log just the fact that the host name part of the email starts with the Google project ID in the control plane).

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Optional;
+import java.util.TreeMap;
+import org.apache.commons.lang3.StringUtils;
+import org.broadinstitute.listener.relay.Utils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class RequestLogger {
```
aherbst-broad, line 16: I converted the HeaderRequestLogger from an `Inspector` to a POJO. The Inspector abstraction was proving to be leaky and not suited to what we need here. Specifically, the ability to inspect the request headers (for cookie + token info), _and_ the response results (status code). 

Rather than stretching that interface and mucking with the chain to create a post-processor abstraction, I opted to make this a POJO and insert a logging statement in the `RelayedHttpRequestProcessor` class. If we find ourselves needing a real chain of command for postprocessing ,we can break out a fancy interface and processing chain similar to what we're doing for requests, but that feels like overkill for now.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
```diff
@@ -0,0 +1,124 @@
...
+      OffsetDateTime requestTimestamp,
+      String prefix)
+      throws IOException, InterruptedException {
+    if (relayedHttpListenerRequest == null) {
+      logger.warn("Null request provided for logging");
+      return;
+    }
+
+    // HTTP headers are case-insensitive
+    var headers = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
```
aherbst-broad, line 43: This is annoying, but as rob rightly pointed out, we are open to case sensitivity issues otherwise.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -16,43 +14,58 @@
...
 public class TokenChecker {
-  private final HttpClient httpClient = HttpClient.newHttpClient();
-  private final String GOOGLE_OAUTH_SERVER =
-      "https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=";
   private final Logger logger = LoggerFactory.getLogger(SamResourceClient.class);
+  private final GoogleTokenInfoClient googleTokenInfoClient;
 
-  public OauthInfo getOauthInfo(String token) throws IOException, InterruptedException {
-    var jwtExpiration = tryDecodeAsB2cToken(token);
+  public TokenChecker(GoogleTokenInfoClient googleTokenInfoClient) {
```
aherbst-broad, line 20: I extracted the "get token info from the google APIs" bits to a separate class to ease testability.

service/src/main/java/org/broadinstitute/listener/relay/OauthInfo.java
```diff
@@ -1,6 +1,7 @@
 package org.broadinstitute.listener.relay;
 
 import java.time.Instant;
+import java.util.Map;
 import java.util.Optional;
 
-public record OauthInfo(Optional<Instant> expiresAt, String error) {}
+public record OauthInfo(Optional<Instant> expiresAt, String error, Map<String, String> claims) {}
```
aherbst-broad, line 7: I am tacking on an extra map of claims here so we can just reach in and get what we need for logging rather than having to break out a separate parameter for each claim.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -16,43 +14,58 @@
...
+        return new OauthInfo(Optional.empty(), googleTokenInfo.error, Map.of());
+      }
+    }
+  }
 
-      if (decoded.expires_in > 0)
-        return new OauthInfo(Optional.of(now.plusSeconds(decoded.expires_in)), decoded.error);
-      else return new OauthInfo(Optional.empty(), decoded.error);
+  private Map<String, String> claimsFromGoogleInfo(GoogleOauthInfoResponse response) {
+    var claims = new HashMap<>(Map.of("sub", response.user_id));
```
aherbst-broad, line 54: I am unsure if this is semantically correct; from what I can tell, the google token info endpoint does not give back a `sub` field, but instead a `user_id` field. Happy to break this out into a separate field if need be.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/HeaderLoggerInspector.java
aherbst-broad: I've refactored a bit and moved where we make the logging calls so we can get both the request data + the status codes.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
rtitle: This makes a lot of sense, feels right to have explicit logging 👍

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
```diff
@@ -0,0 +1,124 @@
...
+    var ua = headers.getOrDefault("User-Agent", "");
+
+    String endpoint = "";
+    if (relayedHttpListenerRequest.getRemoteEndPoint() != null) {
+      endpoint = relayedHttpListenerRequest.getRemoteEndPoint().toString();
+    }
+    var claims = getTokenClaims(headers);
+
+    // log in a single apache-ish line
+    logger.info(
```
rtitle, line 57: All this just goes to stdout right, and is picked up by our collection rules? Was wondering if we need different appenders or something for "apache logs" vs other logs, but maybe not..

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
rtitle: Good point -- I think this is correct to treat `user_id` from google as `sub`. We do this in [apache as well](https://github.com/broadinstitute/terra-helmfile/blob/master/charts/httpd-proxy/templates/_proxy_configmap.yaml#L158).

rtitle (approved):
👍 , nice work and nice improvements along the way re modularity and testability.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/TokenChecker.java
```diff
@@ -16,43 +14,58 @@
...
+      claims.put("email", response.email);
     }
+
+    return claims;
   }
 
-  private Optional<Instant> tryDecodeAsB2cToken(String token) {
+  private Optional<DecodedJWT> tryDecodeAsB2CToken(String token) {
     try {
       var decoded = JWT.decode(token);
```
dinvlad, line 64: I'm not sure this method verifies JWT's signature, but we probably should

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
```diff
@@ -0,0 +1,124 @@
...
+    var headers = new TreeMap<String, String>(String.CASE_INSENSITIVE_ORDER);
+    headers.putAll(relayedHttpListenerRequest.getHeaders());
+
+    var referer = headers.getOrDefault("Referer", "");
+    var origin = headers.getOrDefault("Origin", "");
+    var ua = headers.getOrDefault("User-Agent", "");
+
+    String endpoint = "";
+    if (relayedHttpListenerRequest.getRemoteEndPoint() != null) {
+      endpoint = relayedHttpListenerRequest.getRemoteEndPoint().toString();
```
dinvlad, line 52: Btw `getRemoteEndPoint()` returns _real_ client IP and not Relay's right?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
aherbst-broad: It just goes into the container logs, i.e., you can query for it like so:

```ContainerLog
| where TimeGenerated > ago (48h)
| where ContainerID in ('ad2e08e477db3f9ee91868528344c169bdca9adb5aabfb416ef4e2a67aac5b4c')
| where LogEntry contains "RELAY_REQUEST_RESPONSE"
| order by TimeGenerated desc
| project LogEntry```

**Review by dinvlad — APPROVED** [review]
+1, keeping in mind things we discussed on Slack

File: service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java (hunk 11 above)
Comment (dinvlad) [reply service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java:57]: It might be helpful to have in a JSON form too, but imo that'd blow up log volume unnecessarily, including due to how verbose it is. Workflows uses that in their logs, but I'm not aware of others.

File: service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java [hunk 14]
```diff
@@ -0,0 +1,124 @@
...
+
+    // log in a single apache-ish line
+    logger.info(
+        "{} {} {} \"{}\" [{}] {} \"{} {}\" {} \"{}\" \"{}\" \"{}\"",
+        prefix,
+        endpoint,
+        claims.getOrDefault("sub", ""),
+        claims.getOrDefault("email", ""),
+        requestTimestamp,
+        claims.getOrDefault("idtyp", ""),
```
dinvlad, line 64:
```suggestion
        claims.getOrDefault("idtyp", "-"),
```

service/src/main/java/org/broadinstitute/listener/relay/inspectors/RequestLogger.java
```diff
@@ -0,0 +1,124 @@
...
+
+    String endpoint = "";
+    if (relayedHttpListenerRequest.getRemoteEndPoint() != null) {
+      endpoint = relayedHttpListenerRequest.getRemoteEndPoint().getHostString();
+    }
+    var claims = getTokenClaims(headers);
+
+    // log in a single apache-ish line
+    logger.info(
+        "{} {} {} \"{}\" [{}] {} \"{} {}\" {} \"{}\" \"{}\" \"{}\"",
```
dinvlad, line 58:
```suggestion
        "{} - {} {} \"{}\" \"-\" [{}] {} {} \"{} {} {}\" {} {} \"{}\" \"{}\" \"{}\"",
```
Based on https://github.com/broadinstitute/terra-helmfile/blob/d0d482c72376174d35b773e93e5d54786d0c9dd8/charts/httpd-proxy/templates/_proxy_configmap.yaml#LL18C5-L18C192 
>
```apache
LogFormat "%{X-Forwarded-For}i %l %u \"%{OIDC_CLAIM_email}i\" \"%{X-App-ID}i\" [%{%FT%T}t.%{msec_frac}t%{%z}t] %D %X \"%r\" %>s %b \"%{Referer}i\" \"%{Origin}i\" \"%{User-Agent}i\"" proxy
```
Would be good to plug these:
> %l Remote logname (from identd, if supplied). This will return a dash unless [mod_ident](https://httpd.apache.org/docs/2.4/mod/mod_ident.html) is present and [IdentityCheck](https://httpd.apache.org/docs/2.4/mod/mod_ident.html#identitycheck) is set On.
>
> %D The time taken to serve the request, in microseconds.
>
> %X | Connection status when response is completed:
> X = | Connection aborted before the response completed.
> + = | Connection may be kept alive after the response is sent.
> - = | Connection will be closed after the response is sent.
>
> %r First line of request.
>
I think for `%r` we also log which HTTP version it is, like `GET /status HTTP/1.1`, would be helpful to add that here.

> %b | Size of response in bytes, excluding HTTP headers. In CLF format, i.e. a '-' rather than a 0 when no bytes are sent.

PR #34, opened 2023-05-30
[WOR-1043] Remove cacheable annotation

aherbst-broad:
This [commit](https://github.com/DataBiosphere/terra-azure-relay-listeners/commit/3b326be64f0c7be9ac5155e44dd49b510ff3fa1e) introduced enhanced logging for the azure relay listener. There was also an @ cacheable  annotation errantly included [here](https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/3b326be64f0c7be9ac5155e44dd49b510ff3fa1e/service/src/main/java/org/broadinstitute/listener/relay/inspectors/GoogleTokenInfoClient.java#L18). Since this cache is not configured in the spring app config, the listener fails to startup and relay requests fail. 

This annotation was errantly added during development and not needed afaict. 

Tested in a BEE:
1. Spun up WDS, imported data to a table successfully
2. Spun up a VM, turned on a Jupyter notebook and did a print("Hello world") successfully.
3. Spun up Cromwell, ran a workflow (errored out with config issues unrelated to the listener)


rtitle (approved):
Oh wow.. good find. 👍

PR #35, opened 2023-06-06
[IA-4283] Set dateAccessed from listener properly

mlilynolting:
Leo PR: https://github.com/DataBiosphere/leonardo/pull/3402

Call the correct endpoint on leo (`updateDateAccessed`) and correctly parse the token from the initial request (looking in both Cookie and Authorization headers).

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
@@ -130,6 +130,8 @@ public class SetDateAccessedInspector implements RequestInspector {
       throw new RuntimeException(e);
     }
 
+    logger.debug("Making a call to the last accessed date API at this URL: {}", serviceUrl);
+    
     HttpResponse<String> response;
     try {
       response = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
@@ -110,6 +110,10 @@ public class SetDateAccessedInspector implements RequestInspector {
   }
 
   public void setLastAccessedDateOnService(RelayedHttpListenerRequest relayedHttpListenerRequest) {
+    logger.info("setLastAccessedDateOnService\n\tRelayed request: "
+      .concat(relayedHttpListenerRequestToString(relayedHttpListenerRequest))
+    );
+
     HttpRequest request;
     try {
       request =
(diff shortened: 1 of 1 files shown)


service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -27,7 +27,7 @@ public class SetDateAccessedInspector implements RequestInspector {
   private final int callWindowInSeconds;
   private final HttpClient httpClient;
   private static final int MIN_CALL_WINDOW_IN_SECONDS = 1;
-  private static final String API_ENDPOINT_PATTERN = "%s/api/v2/runtimes/%s/%s/setDateAccessed";
+  private static final String API_ENDPOINT_PATTERN = "%s/api/v2/runtimes/%s/%s/updateDateAccessed";
```
rtitle, line 30: Whoops haha.. good find. 🤦

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -152,4 +150,22 @@ private synchronized boolean hasLastAccessDateExpired() {
   private synchronized void updateLastAccessedDate() {
     lastAccessedDate = lastAccessedDate.plusSeconds(callWindowInSeconds);
   }
+
+  private String httpRequestToString(HttpRequest request) {
```
rtitle, line 154: Doesn't seem like these methods are used?

LizBaldo: Is there a local file missing from this PR that would make use of these new methods? If not I would remove to avoid confusion

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -140,7 +138,7 @@ public void setLastAccessedDateOnService(RelayedHttpListenerRequest relayedHttpL
       throw new RuntimeException(e);
     }
 
-    logger.info(
+    logger.debug(
```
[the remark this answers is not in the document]
mlilynolting: My mistake; thought this was the original state.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
mlilynolting: Ha! Purely here for debug purposes. Thanks

LizBaldo (approved):
LGTM, thanks :)

jdcanas (approved):

PR #40, opened 2023-06-27
[IA-4401] Allow get status endpoints without updating dateAccessed

mlilynolting:
### What
- Allow api/status and welder/status endpoints to pass through the relay listener without touching dateAccessed on the runtime (which keeps the runtime from pausing).

### Why
-

### Testing strategy
- [ ]


mlilynolting:
weird discrepancy between local and remote.

PR #41, opened 2023-06-29
[IA-4404] isValidOrigin - Added unit test and updated logic.

n8landolt:
### What
- Added functionality to allow port to be part of the URL schema. Added a test to verify.

### Why
- Requests were being denied if they were from localhost:3000

### Testing strategy
- [ ]

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -53,11 +53,10 @@ public class Utils {
   }
 
   public static boolean isValidOrigin(String origin, CorsSupportProperties corsSupportProperties) {
-    if (corsSupportProperties.validHosts().contains("*")) {
+    if (origin.isEmpty() || corsSupportProperties.validHosts().contains("*")) {
       return true;
     }
 
-    String stringToCheck;
     // We want to strip the protocol.
     URL url;
     try {
@@ -67,12 +66,11 @@ public class Utils {
       return false;
     }
 
-    return origin.isEmpty()
-        || corsSupportProperties.validHosts().stream()
-            .anyMatch(
-                validHost ->
-                    // We still need to strip out spaces!
-                    validHost.replace(" ", "").equals(url.getAuthority()));
+    return corsSupportProperties.validHosts().stream()
+        .anyMatch(
+            validHost ->
+                // We still need to strip out spaces!
+                validHost.replace(" ", "").equals(url.getAuthority()));
   }
 
   public static Optional<String> getToken(Map<String, String> headers) {
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -62,15 +62,12 @@ public class Utils {
     try {
       url = new URL(origin);
     } catch (MalformedURLException e) {
-      logger.error(String.format("Error parsing URL:%s. MalformedURLException: %s", origin, e.getMessage()));
+      logger.error(
+          String.format("Error parsing URL:%s. MalformedURLException: %s", origin, e.getMessage()));
       return false;
     }
 
-    return corsSupportProperties.validHosts().stream()
-        .anyMatch(
-            validHost ->
-                // We still need to strip out spaces!
-                validHost.trim().equals(url.getAuthority()));
+    return corsSupportProperties.validHosts().contains(url.getAuthority());
   }
 
   public static Optional<String> getToken(Map<String, String> headers) {
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -62,7 +62,7 @@ public class Utils {
     try {
       url = new URL(origin);
     } catch (MalformedURLException e) {
-      logger.info(e.toString());
+      logger.error(String.format("Error parsing URL:%s. MalformedURLException: %s", origin, e.getMessage()));
       return false;
     }
 
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -70,7 +70,7 @@ public class Utils {
         .anyMatch(
             validHost ->
                 // We still need to strip out spaces!
-                validHost.replace(" ", "").equals(url.getAuthority()));
+                validHost.trim().equals(url.getAuthority()));
   }
 
   public static Optional<String> getToken(Map<String, String> headers) {
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -8,7 +8,6 @@ import java.net.URL;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Optional;
-import java.util.regex.Pattern;
 import org.broadinstitute.listener.config.CorsSupportProperties;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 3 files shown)


service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -69,11 +74,11 @@ public static boolean isValidOrigin(String origin, CorsSupportProperties corsSup
     String finalStringToCheck = stringToCheck;
     return origin.isEmpty()
         || corsSupportProperties.validHosts().stream()
-        .anyMatch(
-            validHost ->
-                Pattern.matches(
-                    validHost.replace(".", "\\.").replace("*", ".*").replace(" ", ""),
-                    finalStringToCheck));
+            .anyMatch(
```
mlilynolting, line 77: None of our refererConfigs in any environment I'm aware of actually use wildcard matching other than the single wildcard alone (`"*"`), and if included it will break Leonardo (pending https://broadworkbench.atlassian.net/browse/IA-4368). I think that this should be removed and the solo wildcard handled specifically.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -58,7 +58,12 @@ public static boolean isValidOrigin(String origin, CorsSupportProperties corsSup
     // We want to strip the protocol.
     try {
       URL url = new URL(origin);
-      stringToCheck = url.getHost();
```
mlilynolting, line 61: Replacing with `url.getAuthority()` will include both host and port (if specified), as well as omitting the trailing slash.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -58,7 +58,12 @@ public static boolean isValidOrigin(String origin, CorsSupportProperties corsSup
...
+      if (url.getPort() != -1) {
+        stringToCheck = url.getHost() + ':' + url.getPort();
+      } else {
+        stringToCheck = url.getHost();
+      }
+      logger.info(url.getPath());
     } catch (MalformedURLException e) {
       stringToCheck = origin;
```
mlilynolting, line 68: If we handled the wildcard case and the empty case before the try, we'd be able to simply fail out from here, right?

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -69,11 +74,11 @@ public static boolean isValidOrigin(String origin, CorsSupportProperties corsSup
     String finalStringToCheck = stringToCheck;
```
mlilynolting, line 74: Commenting here because GH won't let me comment on line 66 (idk): can we demote these to debug logs once validated? They're super spammy.

mlilynolting (changes requested):

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
n8landolt: Oh? Let me try!

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -54,26 +53,26 @@ public static void writeCORSHeaders(
...
     String stringToCheck;
     // We want to strip the protocol.
+    URL url;
     try {
-      URL url = new URL(origin);
-      stringToCheck = url.getHost();
+      url = new URL(origin);
     } catch (MalformedURLException e) {
-      stringToCheck = origin;
+      logger.info(e.toString());
```
mlilynolting, line 66: Would `logger.error` or `warn` be more appropriate here?

mlilynolting: Oh and can we get a wrapped bit of error text, like "Error parsing origin URL:" ? Will make it easier to search for if we ever need to debug from logs

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -54,26 +53,26 @@ public static void writeCORSHeaders(
...
+      return true;
+    }
+
     String stringToCheck;
     // We want to strip the protocol.
+    URL url;
     try {
-      URL url = new URL(origin);
-      stringToCheck = url.getHost();
+      url = new URL(origin);
```
mlilynolting, line 64: Quick check - if origin is empty, will this trigger the MalformedURLException? We don't want to, in that case, right? Should we add the empty check above, where we check the wildcard?

service/src/test/java/org/broadinstitute/listener/relay/UtilTest.java
mlilynolting, line 1: Is there an empty test? An invalid origin test? And did the wildcard allowAll get removed or is it just somewhere else?

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -54,26 +53,24 @@ public static void writeCORSHeaders(
...
-    return origin.isEmpty()
-        || corsSupportProperties.validHosts().stream()
+    return corsSupportProperties.validHosts().stream()
         .anyMatch(
             validHost ->
-                Pattern.matches(
-                    validHost.replace(".", "\\.").replace("*", ".*").replace(" ", ""),
-                    finalStringToCheck));
+                // We still need to strip out spaces!
+                validHost.replace(" ", "").equals(url.getAuthority()));
```
mlilynolting, line 73: can use `validHost.trim()` here

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -54,26 +53,24 @@ public static void writeCORSHeaders(
...
-    return origin.isEmpty()
-        || corsSupportProperties.validHosts().stream()
+    return corsSupportProperties.validHosts().stream()
         .anyMatch(
             validHost ->
-                Pattern.matches(
-                    validHost.replace(".", "\\.").replace("*", ".*").replace(" ", ""),
-                    finalStringToCheck));
+                // We still need to strip out spaces!
+                validHost.trim().equals(url.getAuthority()));
```
mlilynolting, line 73: One last thing, and then I swear to God I'm done😅 - the `trim()` feels out of place here, the more I think about it. We could normalize the validHosts at the source, inside the CorsSupportProperties record, with an explicit canonical constructor: e.g.
```
public record CorsSupportProperties (
    String preflightMethods,
    String allowHeaders,
    String maxAge,
    String contentSecurityPolicy,
    List<String> validHosts) {
				public CorsSupportProperties (
				    String preflightMethods,
				    String allowHeaders,
    				String maxAge,
    				String contentSecurityPolicy,
    				List<String> validHosts) {
    						this.preflightMethods = preflightMethods;
    						this.allowHeaders = allowHeaders;
    						this.maxAge = maxAge;
    						this.contentSecurityPolicy = contentSecurityPolicy;
    						this.validHosts = validHosts.stream().map(host -> host.trim()).collect(Collectors.toList())
				}
}
```
I know this is 12 lines in the model to replace a single method call in the service, but it feels better to normalize this data which is used all over.

mlilynolting: (forgive if any of this is wrong; I'm not super familiar with streams or records, having worked mostly in older Java. But they're very cool)

mlilynolting (commented):
Thanks for addressing everything! My last ask is more a code smell nit than a strong requirement, so if you think it would be trouble to implement let me know.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
n8landolt: This is an excellent suggestion! Thank you. I'll make the adjustments

mlilynolting (approved):

PR #42, opened 2023-06-30
[IA-4401] Allow get status endpoints without updating dateAccessed

mlilynolting:
### What
- Filter requests which pass through SetDateAccessedInspector: if request includes the custom HTTP header `X-SetDateAccessedInspector-Action=ignore`, do not send a PATCH updateDateAccessed. This allows Leonardo to exempt its metrics and monitoring calls from this behavior, preventing a loop where Leo's checks for runtime health prevent the runtime from autopausing.

### Why
- Prevents Leo's MetricsMonitor from blocking Azure runtime autopause

### Testing strategy
- Create an Azure runtime with autopause >25m
- Verify it stops after that length of time idle
- Create apps (RStudio, Cromwell)
- Verify these function normally

- [x] I validated these changes on my BEE:
* autopause works with a DS runtime
* autopause works with an FS runtime
* autopause works with and without Cromwell running
* CRUD operations work on Azure runtimes
* autopause works on GCP runtimes (Jupyter, RStudio)
* CRUD operations work on GCP runtimes

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -19,8 +19,15 @@ public class Utils {
   public static final Logger logger = LoggerFactory.getLogger(Utils.class);
   public static final String TOKEN_NAME = "LeoToken";
   public static final String SET_COOKIE_API_PATH = "setcookie";
-  public static final List<String> GET_STATUS_API_PATH = List.of("api", "status");
-  public static final List<String> GET_WELDER_STATUS_API_PATH = List.of("welder", "status");
+
+  /**
+   * Paths which should not trigger dateAccessed runtime updates. TODO: move related logic into
+   * SetDateAccessed* classes.
+   */
+  public static final List<List<String>> KEEP_ALIVE_EXEMPT_PATHS =
+      List.of(List.of("api", "status"), List.of("welder", "status"));
+
+  private static final int PATH_PREFIX_LENGTH = 2;
 
   public static final Optional<String> getTokenFromAuthorization(Map<String, String> headers) {
     var authValue = headers.getOrDefault(AUTHORIZATION, null);
@@ -43,20 +50,23 @@ public class Utils {
    * Whether the request represents user or kernel activity, and should thus trigger a touch of the
    * runtime to prevent autopause.
    */
-  public static boolean isNotKeepAliveRequest(RelayedHttpListenerRequest listenerRequest) {
-    var isGetStatusRequest =
-        listenerRequest.getHttpMethod().equals("GET") && isGetStatusPath(listenerRequest.getUri());
-    return !isGetStatusRequest;
-  }
-
-  private static boolean isGetStatusPath(URI uri) {
+  public static boolean isKeepAliveRequest(RelayedHttpListenerRequest listenerRequest) {
+    URI uri = listenerRequest.getUri();
     var splitted = Arrays.asList(uri.getPath().split("/"));
-    if (splitted.size() == 4) {
-      var tail = splitted.subList(splitted.size() - 2, splitted.size());
-      return GET_STATUS_API_PATH.equals(tail) || GET_WELDER_STATUS_API_PATH.equals(tail);
-    } else {
-      return false;
-    }
+    boolean isKeepAliveExempt =
+        listenerRequest.getHttpMethod().equals("GET")
+            && KEEP_ALIVE_EXEMPT_PATHS.stream()
+                .anyMatch(
+                    exemptPath -> {
+                      int exemptPathSize = exemptPath.size();
+                      if (splitted.size() != PATH_PREFIX_LENGTH + exemptPathSize) {
+                        return false;
+                      }
+                      var tail =
+                          splitted.subList(splitted.size() - exemptPathSize, splitted.size());
+                      return exemptPath.equals(tail);
+                    });
+    return !isKeepAliveExempt;
   }
 
   public static void writeCORSHeaders(
--- a/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/Utils.java
@@ -39,7 +39,11 @@ public class Utils {
     }
   }
 
-  public static boolean isNotGetStatusRequest(RelayedHttpListenerRequest listenerRequest) {
+  /**
+   * Whether the request represents user or kernel activity, and should thus trigger a touch of the
+   * runtime to prevent autopause.
+   */
+  public static boolean isNotKeepAliveRequest(RelayedHttpListenerRequest listenerRequest) {
     var isGetStatusRequest =
         listenerRequest.getHttpMethod().equals("GET") && isGetStatusPath(listenerRequest.getUri());
     return !isGetStatusRequest;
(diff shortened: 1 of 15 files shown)


service/src/test/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessorTest.java
```diff
@@ -83,10 +83,10 @@ void setUp() {
     targetResponseHeaders.put("RES_HEADER", List.of("RES_VALUE"));
     requestHeaders = new HashMap<>();
     requestHeaders.put("REQ_HEADER", "REQ_VALUE");
-    requestHeaders.put("Origin", "app.terra.bio");
+    requestHeaders.put("Origin", "http://app.terra.bio");
 
     requestHeaders_invalidOrigin = new HashMap<>();
-    requestHeaders_invalidOrigin.put("Origin", "malicious.website.com");
+    requestHeaders_invalidOrigin.put("Origin", "http://malicious.website.com");
```
mlilynolting, line 89: Origin header has scheme. ValidHosts does not.

mlilynolting:
Jupyter starts! :D

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -18,6 +20,15 @@ public class Utils {
   public static final String TOKEN_NAME = "LeoToken";
   public static final String SET_COOKIE_API_PATH = "setcookie";
 
+  /**
+   * Paths which should not trigger dateAccessed runtime updates. TODO: move related logic into
+   * SetDateAccessed* classes.
```
mlilynolting, line 25: I would love to do this (encapsulate logic in the SetDateAccessedListener instead of leaving it hanging out in Utils).

I would also like to make the paths configurable and have Leo pass them in.

I don't think these beautifications should block this critical bugfix, but maybe they can be addressed when we add a new non-keep-alive path to the runtime metrics monitor.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -18,6 +20,15 @@ public class Utils {
...
+   */
+  public static final List<List<String>> KEEP_ALIVE_EXEMPT_PATHS =
+      List.of(List.of("api", "status"), List.of("welder", "status"));
+
+  private static final int PATH_PREFIX_LENGTH = 2;
```
mlilynolting, line 30: Realizing now that this design is also quite brittle - I think if I were doing it right the KEEP_ALIVE_EXEMPT_PATHS might be regex patterns, like `List.of(".*",".*","api","status")` which could then be directly compared to the splitted uri path.

service/src/main/java/org/broadinstitute/listener/relay/Utils.java
```diff
@@ -35,6 +46,29 @@ public static boolean isSetCookiePath(URI uri) {
...
+  public static boolean isKeepAliveRequest(RelayedHttpListenerRequest listenerRequest) {
+    URI uri = listenerRequest.getUri();
+    var splitted = Arrays.asList(uri.getPath().split("/"));
+    boolean isKeepAliveExempt =
+        listenerRequest.getHttpMethod().equals("GET")
+            && KEEP_ALIVE_EXEMPT_PATHS.stream()
+                .anyMatch(
+                    exemptPath -> {
+                      int exemptPathSize = exemptPath.size();
+                      if (splitted.size() != PATH_PREFIX_LENGTH + exemptPathSize) {
```
LizBaldo, line 62: Oh yeah, this does feel a bit brittle indeed, using regex patterns would be best here, if possible i would do it in this PR

LizBaldo (commented):
Nice work! I agree that the immediate follow up step would be to handle this exception list in leonardo, but not as part of this PR.

I do agree that making the path matching logic a bit more robust by using Regex would be best here though

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -88,10 +102,49 @@ public boolean inspectWebSocketUpgradeRequest(
...
 
+  /**
+   * Inspect the request, calling Leonardo to updateDateAccessed on our resource if the request
+   * represents an action from a user other than the Leonardo service account.
+   */
   @Override
   public boolean inspectRelayedHttpRequest(RelayedHttpListenerRequest relayedHttpListenerRequest) {
+    logger.warn("DATE ACCESSED LISTENER !!! {}", relayedHttpListenerRequest.getUri());
+    logger.warn(relayedHttpListenerRequest.getHeaders().toString());
+    if (isLeonardoServiceAccountUser(relayedHttpListenerRequest)) {
```
giventocode, line 113: I see the overall approach. However, I am wondering if relying on an email address from the claim to identify what requests to ignore is a bit brittle. As an alternative, what about relying on the host name of caller - with a list in the config of what look for? Or filter by a custom header (e.g.  leo-status-check) that Leo can set when making status checks.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
```diff
@@ -88,10 +91,32 @@ public boolean inspectWebSocketUpgradeRequest(
...
+   * Inspect the request. Call Leonardo to updateDateAccessed on our resource, unless the request
+   * headers include {@code X-SetDateAccessedInspector-Action=ignore}.
+   *
+   * @return whether to relay the original request.
+   */
   @Override
   public boolean inspectRelayedHttpRequest(RelayedHttpListenerRequest relayedHttpListenerRequest) {
+    if (isActionIgnore(relayedHttpListenerRequest)) {
+      logger.info("SetDateAccessedInspector will ignore this request.");
+      return true;
```
LizBaldo, line 104: Might be a java newbie question, but why return true here?

LizBaldo (approved):
Oh I see what you did there, great job! Very nice/elegant to have leonardo specifically tell the listener to ignore the jupyter proxy / welder requests 👍 

Before you merge, could you edit the PRs descriptions with this new method now?

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
mlilynolting: `inspectRelayedHttpRequest()` returns a bool which is used by the inspectors pipeline: all inspectors must return `true` in order for the request to be forwarded to the target. `SetDateAccessedInspector` never impedes requests; its action is either a no-op (ignore, this case) or a message to Leo to updateDateAccessed (the other case). In both cases, returning `true` just OKs the request to proceed.

mlilynolting: I could have moved the boolean return outside of the if-else block, since it should always return true, but having it explicit for each case seemed slightly more readable to a future dev (who might otherwise see the ignore log and think that an early-return might have been overlooked).

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SetDateAccessedInspector.java
mlilynolting: Done! @giventocode thank you for all your feedback :)

PR #43, opened 2023-07-12
AJ-1147: cache Sam responses much less aggressively

davidangb:
### Jira Ticket: https://broadworkbench.atlassian.net/browse/AJ-1147

**_Reviewer_**: the latest version of this PR reduces the cache TTL to 90 seconds. What is your opinion of an appropriate TTL? Should we disable the cache entirely?

## Summary of changes:

Change the Sam caching eviction strategy:
* 1 hr TTL -> 90 second TTL
* expireAfterAccess -> expireAfterWrite

The expireAfterAccess -> expireAfterWrite change means that Sam responses "should be automatically removed from the cache once a fixed duration has elapsed after the entry’s creation, or the most recent replacement of its value" instead of being "automatically removed from the cache once a fixed duration has elapsed after the entry’s creation, the most recent replacement of its value, or its last read." (reference https://camel.apache.org/components/3.20.x/caffeine-cache-component.html#_component_options)

In other words, we expire Sam responses 90 seconds after the Listener receives them, instead of 1 hour after the last time the cached response was read.

### Why

With the previous Listener caching strategy, it is quite easy to hold on to a cached value long after Terra has updated the underlying permissions.

### Testing strategy
- [ ]

### DIFF
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -5,7 +5,7 @@ spring:
   application.name: terra-azure-relay-listener
   cache:
     cache-names: expiresAt
-    caffeine.spec: maximumSize=100,expireAfterWrite=600s
+    caffeine.spec: maximumSize=100,expireAfterWrite=300s
 
 listener:
   relayConnectionString:
--- a/service/src/main/resources/application.yml
+++ b/service/src/main/resources/application.yml
@@ -5,7 +5,7 @@ spring:
   application.name: terra-azure-relay-listener
   cache:
     cache-names: expiresAt
-    caffeine.spec: maximumSize=100,expireAfterWrite=300s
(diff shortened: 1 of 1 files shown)


LizBaldo (approved):
LGTM! The only way to test it now would be to publish the new listener image, have a PR in leo with that new image hash and then test on a BEE (aka a full E2E test 😬)

LizBaldo (approved):
Do you need support getting this into Leo for testing @davidangb ?

davidangb:
@LizBaldo I have tested this in a bee (via https://github.com/DataBiosphere/leonardo/pull/3493), and it works as expected. I don't need logistical help for getting this though - just need opinions on the appropriate cache TTL or if we should have a cache at all!

ashanhol (approved):

tlangs:
I'm good with 90 seconds!

PR #45, opened 2023-07-29
AJ-1083: Add liveness and readiness checks via Spring actuator

rtitle:
## Summary of changes:
Enable [Spring Boot Actuator](https://www.baeldung.com/spring-liveness-readiness-probes) to set up liveness and readiness probes in Spring Boot. These are configured in the Relay Listener helm chart in: https://github.com/broadinstitute/terra-helmfile/pull/4411.

The Spring Boot endpoints in the relay listener are not exposed externally. Additionally I added a `/listenerstatus` endpoint through the Relay Namespace, which also returns the actuator health endpoint payload.

### Testing strategy
Deployed via a BEE. Verified:
1. Listener pods come up, k8s events show liveness/readiness working
2. Listener logs show liveness/readiness state changes
3. `/listenerstatus` endpoint works
4. Leo metrics are populated with listener health

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
@@ -8,6 +8,10 @@ import org.springframework.boot.availability.ReadinessState;
 import org.springframework.context.event.EventListener;
 import org.springframework.stereotype.Component;
 
+/**
+ * Listens for availability change events from Spring Boot Actuator and writes log messages.
+ * See: https://www.baeldung.com/spring-boot-actuators
+ */
 @Component
 public class AvailabilityListener {
   private static final Logger logger = LoggerFactory.getLogger(AvailabilityListener.class);
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
@@ -19,7 +19,7 @@ public class AvailabilityListener {
         logger.error("LivenessState: {}", event.getState());
         break;
       case CORRECT:
-        logger.debug("LivenessState: {}", event.getState());
+        logger.info("LivenessState: {}", event.getState());
         break;
       default:
         logger.error("LivenessState {} is not a known state", event.getState());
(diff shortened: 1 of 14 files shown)


rtitle:
Adding a few reviewers -- I think this is working well in a BEE. Feedback welcome.

service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -194,6 +207,47 @@ public Result writeSetCookieResponse(RelayedHttpListenerContext context) {
...
+          String.format(
+              "Origin %s not allowed. Error Code: RHRP-002",
+              requestHeaders.getOrDefault("Origin", "")));
+      return Result.FAILURE;
+    }
+
+    RelayedHttpListenerResponse listenerResponse = context.getResponse();
+
+    // Write headers
+    Utils.writeCORSHeaders(
```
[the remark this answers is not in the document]
rtitle: Good point, CORS probably not needed here. Also I thought I'd check `isValidOrigin` to be conservative, but it's probably also not needed -- all our status endpoints are public anyway. I can remove this.

service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
```diff
@@ -0,0 +1,42 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.availability.AvailabilityChangeEvent;
+import org.springframework.boot.availability.LivenessState;
+import org.springframework.boot.availability.ReadinessState;
+import org.springframework.context.event.EventListener;
+import org.springframework.stereotype.Component;
+
+@Component
```
LizBaldo, line 11: Just a thought, but reading the article you linked in the PR description was very useful for me, I think it would be a good idea to link it in the code here as a comment as well https://www.baeldung.com/spring-liveness-readiness-probes

LizBaldo (approved):

service/src/main/java/org/broadinstitute/listener/relay/http/AvailabilityListener.java
rtitle: Good idea! Will do.

ashanhol (approved):

ashanhol:
You might wanna look at the code smells

rtitle:
> You might wanna look at the code smells

Thanks, fixed all! A lot of those are old habits which are "smells" nowadays 👃

PR #47, opened 2023-08-11
[WM-2157] Additional check for workspace access

salonishah11:
### What
- Check whether a user is in the right workspace when trying to access an app or notebook

### Why
- When app owner is removed from a workspace, they lose access to their app APIs

### Testing strategy
Tested it in a BEE by updating the Relay Listener image Leo points to. Testing steps:
- User A creates a workspace and Cromwell app
- User A is able to see Cromwell app and access its APIs
- User A adds User B as owner on the workspace
- User B removes User A from the workspace
- User A can no longer access the workspace or Cromwell app APIs. Relay listener returns 403 on the request:
```
% curl -i -X 'GET' \
  'https://lz1...../cbas/api/batch/v1/methods?show_versions=false' \
  -H 'accept: application/json' \
  -H 'Authorization: Bearer <redacted>'
HTTP/1.1 403 The listener rejected the request. Tracking ID:2aa...
Via: 1.1 lz1c....9e5.servicebus.windows.net
Content-Length: 0
Server: Microsoft-HTTPAPI/2.0
Strict-Transport-Security: max-age=31536000
Date: Thu, 17 Aug 2023 20:31:10 GMT
```
Logs in Relay listener show "Unauthorized request. User doesn't have access to workspace." being logged when this request is made:
```
2023-08-17 20:31:10.458   INFO 1 --- [t@376926745-576] com.microsoft.azure.relay.RelayLogger    : HybridHttpConnection(TrackingId:2aa..., Address:https://lz1c...e5.servicebus.windows.net): request initializing.
2023-08-17 20:31:10.459   INFO 1 --- [     Thread-171] com.microsoft.azure.relay.RelayLogger    : HybridHttpConnection(TrackingId:2aa..., Address:https://lz1c...9e5.servicebus.windows.net): Request method: GET.
2023-08-17 20:31:10.459   INFO 1 --- [     Thread-171] com.microsoft.azure.relay.RelayLogger    : HybridHttpConnection(TrackingId:2aa..., Address:https://lz1c5...e5.servicebus.windows.net): Invoking user RequestHandler.
2023-08-17 20:31:10.638  ERROR 1 --- [undedElastic-59] o.b.l.r.inspectors.SamResourceClient     : Unauthorized request. User doesn't have access to workspace.
2023-08-17 20:31:10.638   INFO 1 --- [undedElastic-59] o.b.l.r.inspectors.InspectorsProcessor   : Inspection result for the HTTP request. Result: false, URI:sb://lz1c...<path-to-app>..ca34f/cbas/api/batch/v1/methods?show_versions=false
2023-08-17 20:31:10.638   INFO 1 --- [undedElastic-59] com.microsoft.azure.relay.RelayLogger    : HybridHttpConnection(TrackingId:2aa..., Address:https://lz1c5...9e5.servicebus.windows.net)+ResponseStream is closing.
2023-08-17 20:31:10.638   INFO 1 --- [undedElastic-59] com.microsoft.azure.relay.RelayLogger    : HybridHttpConnection(TrackingId:2aa..., Address:https://lz1c...9e5.servicebus.windows.net): Sending the response command on the control connection, status: 403.
2023-08-17 20:31:10.638   INFO 1 --- [undedElastic-59] com.microsoft.azure.relay.RelayLogger    : com.microsoft.azure.relay.HybridConnectionListener$ControlConnection@637c0751: Response command was sent: {"response":{"requestId":"2aa...","body":false,"statusCode":403,"statusDescription":"The listener rejected the request. Tracking ID:2aa..."}}
```

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
@@ -45,6 +45,7 @@ public class AppConfiguration {
     samClient.setBasePath(properties.getSamInspectorProperties().samUrl());
     // return a simple resolver that uses the configuration value.
     return new SamResourceClient(
+        properties.getSetDateAccessedInspectorProperties().workspaceId(),
         properties.getSamInspectorProperties().samResourceId(),
         properties.getSamInspectorProperties().samResourceType(),
         samClient,
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
@@ -41,6 +41,13 @@ public class SamResourceClient {
         samClient.setAccessToken(accessToken);
         var resourceApi = new ResourcesApi(samClient);
 
+        // check that user has access to workspace
+        var workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
+        if (!workspaceAccess) {
+          logger.error("Unauthorized request. User doesn't have access to workspace");
+          return Instant.EPOCH;
+        }
+
         var res = resourceApi.resourcePermissionV2(samResourceType, samResourceId, samAction);
         if (res) return oauthInfo.expiresAt().get();
         else {
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
@@ -2,6 +2,7 @@ package org.broadinstitute.listener.relay.inspectors;
 
 import java.io.IOException;
 import java.time.Instant;
+import java.util.UUID;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.broadinstitute.dsde.workbench.client.sam.api.ResourcesApi;
@@ -10,6 +11,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.cache.annotation.Cacheable;
 
 public class SamResourceClient {
+  private final UUID workspaceId;
   private final String samResourceId;
   private final String samResourceType;
   private final TokenChecker tokenChecker;
@@ -19,11 +21,13 @@ public class SamResourceClient {
   private final Logger logger = LoggerFactory.getLogger(SamResourceClient.class);
 
   public SamResourceClient(
+      UUID workspaceId,
       String samResourceId,
       String samResourceType,
       ApiClient samClient,
       TokenChecker tokenChecker,
       String samAction) {
+    this.workspaceId = workspaceId;
     this.samResourceId = samResourceId;
     this.samResourceType = samResourceType;
     this.tokenChecker = tokenChecker;
@@ -42,9 +46,10 @@ public class SamResourceClient {
         var resourceApi = new ResourcesApi(samClient);
 
         // check that user has access to workspace
-        boolean workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
+        boolean workspaceAccess =
+            resourceApi.resourcePermissionV2("workspace", workspaceId.toString(), "read");
         if (!workspaceAccess) {
-          logger.error("Unauthorized request. User doesn't have access to workspace");
+          logger.error("Unauthorized request. User doesn't have access to workspace.");
           return Instant.EPOCH;
         }
 
--- a/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
@@ -42,7 +42,7 @@ public class SamResourceClient {
         var resourceApi = new ResourcesApi(samClient);
 
         // check that user has access to workspace
-        var workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
+        boolean workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
         if (!workspaceAccess) {
           logger.error("Unauthorized request. User doesn't have access to workspace");
           return Instant.EPOCH;
--- a/service/src/test/java/org/broadinstitute/listener/relay/inspectors/SamResourceClientTest.java
+++ b/service/src/test/java/org/broadinstitute/listener/relay/inspectors/SamResourceClientTest.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.time.Instant;
 import java.util.Map;
 import java.util.Optional;
+import java.util.UUID;
 import org.broadinstitute.dsde.workbench.client.sam.ApiClient;
 import org.broadinstitute.dsde.workbench.client.sam.ApiException;
 import org.broadinstitute.dsde.workbench.client.sam.ApiResponse;
@@ -30,7 +31,8 @@ class SamResourceClientTest {
   @BeforeEach
   void setUp() throws IOException, InterruptedException, ApiException {
     samResourceClient =
-        new SamResourceClient("resourceId", "resourceType", apiClient, tokenChecker, "myaction");
+        new SamResourceClient(
+            UUID.randomUUID(), "resourceId", "resourceType", apiClient, tokenChecker, "myaction");
   }
 
   @Test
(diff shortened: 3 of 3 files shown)


service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -41,6 +41,13 @@ public Instant checkPermission(String accessToken) {
         samClient.setAccessToken(accessToken);
         var resourceApi = new ResourcesApi(samClient);
 
+        // check that user has access to workspace
+        var workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
```
rtitle: I think `samResourceId` is wrong -- that would be the ID of the runtime/kubernetes-app, not the workspace.

You might need to add a new contructor-arg to this class and pass it from `AppConfiguration`

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -41,6 +41,13 @@ public Instant checkPermission(String accessToken) {
         samClient.setAccessToken(accessToken);
         var resourceApi = new ResourcesApi(samClient);
 
+        // check that user has access to workspace
+        boolean workspaceAccess = resourceApi.resourcePermissionV2("workspace", samResourceId, "read");
```
rtitle, line 45: I think `samResourceId` is wrong -- that would be the ID of the runtime/kubernetes-app, not the workspace.

You might need to add a new contructor-arg to this class and pass it from AppConfiguration (which should have the the workspaceId for other things).

salonishah11: Ah I see. It looks like `setDateAccessedInspectorProperties` in application config has workspace ID. I will use that to pass it to SamResourceClient. Thank you for pointing this out!

rtitle (approved):
LGTM 👍

rtitle:
FYI - I just merged https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/48 so you may need to rebase this.

service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
```diff
@@ -41,6 +41,7 @@ public TargetResolver targetResolver() {
   @Bean
   public SamResourceClient samResourceClient(TokenChecker tokenChecker) {
     return new SamResourceClient(
+        properties.getSetDateAccessedInspectorProperties().workspaceId(),
```
mlilynolting, line 44: I think this might be meant to be `getSamInspectorProperties()`!

mlilynolting (commented):
Looks good barring one thing which I think is a typo :)

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -46,6 +50,14 @@ public Instant checkPermission(String accessToken) {
         var apiClient = getApiClient(accessToken);
         var resourceApi = new ResourcesApi(apiClient);
 
+        // check that user has access to workspace
+        boolean workspaceAccess =
+            resourceApi.resourcePermissionV2("workspace", workspaceId.toString(), "read");
+        if (!workspaceAccess) {
+          logger.error("Unauthorized request. User doesn't have access to workspace.");
+          return Instant.EPOCH;
+        }
```
davidangb, line 59: I won't block the PR for this - but it would be great if this was optional and controlled by a value sent by Leo. This extra Sam call is extraneous for `kubernetes-app-shared` apps, e.g. WDS. It won't break anything, but it's an extra API call that could add latency and/or the chance of transient failures.

service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
salonishah11: I had initially thought of the same property to use. But it seems that based on [README](https://github.com/DataBiosphere/terra-azure-relay-listeners#configuration-properties-1) and [application.yaml](https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/4ddf0352e890a6c794cc8835056265500205ab46/service/src/main/resources/application.yml#L30-L34) `samInspectorProperties` doesn't have a workspace id in it. The only config that had the workspace id was [setDateAccessedInspectorProperties](https://github.com/DataBiosphere/terra-azure-relay-listeners/blob/4ddf0352e890a6c794cc8835056265500205ab46/service/src/main/resources/application.yml#L24-L28).

mlilynolting: Hm, ok. Tech debt :(

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
mlilynolting: Is this something that would apply statically to the listeners instance? Like, WDS has a listeners instance, and it could safely ignore all workspace access checks?

mlilynolting (approved):

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
davidangb: Exactly - the listener deployed for WDS could skip this check; the listener deployed for Cromwell still needs the check. When CBAS moves from single-user to multi-user it could also skip this check.

mlilynolting: we could tidy two birds with one stone and update leonardo to send a new samChecker parameter `workspaceId`, which if specified would do this check

mlilynolting: I am ok with calling this out of scope for this bug ticket but I won't speak for WDS as I'm not familiar

salonishah11: For WDS would it be setting a config value in Relay Listener that Leonardo will set? I am not very familiar how and when this happens

mlilynolting: YMMV but,,

You'd probably be making changes around here: https://github.com/DataBiosphere/leonardo/blob/84549cc1f5571f867f087b9e5c39de0b267e9670/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L169

That's where the configuration is applied to the terra-azure-relay-listeners instance for a new AKS app. You'd want to check if it's a shared app, which I think is what's happening for example here: https://github.com/DataBiosphere/leonardo/blob/84549cc1f5571f867f087b9e5c39de0b267e9670/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AKSInterpreter.scala#L413

Otherwise you could set the new field in the helmchart values, as well as in AzurePubSubHandler which creates the listener for new runtimes: https://github.com/DataBiosphere/leonardo/blob/84549cc1f5571f867f087b9e5c39de0b267e9670/http/src/main/scala/org/broadinstitute/dsde/workbench/leonardo/util/AzurePubsubHandler.scala#L130 (and a corresponding change in azure_vm_init_script)

salonishah11: Would it make sense to do that in a separate ticket instead of this ticket? It might require more investigating and changes in more places than Relay Listener

mlilynolting: Feels ok to me to have this be a first pass with a followup ticket to refine.

rtitle: Is the workspace check actually extraneous for shared apps? What about this scenario:
1. I create a shared app in a workspace
2. I leave the workspace
3. I try to access my shared app

Would we need to check the workspace to effectively block (3)?

davidangb: > Would we need to check the workspace to effectively block (3)

This is exactly the scenario I validated in [AJ-1147](https://broadworkbench.atlassian.net/browse/AJ-1147). Since `kubernetes-app-shared` roles are inherited from `workspace` roles - and ONLY inherited, there are no direct roles - this works as expected. See also https://github.com/DataBiosphere/leonardo/pull/3496

rtitle: Ah I see, cool

PR #48, opened 2023-08-12
IA-4489: setCookie endpoint should check user enablement and ToS

rtitle:
JIRA: https://broadworkbench.atlassian.net/browse/IA-4489

### What
- Add `SamResourceClient.isUserEnabled` method and call it from the `setCookie` endpoint
- Added unit test coverage for `setCookie`
- Improvements to `SamResourceClient` -- instantiate a new client for each request but use shared thread pool

### Why
- During security review, determined that `setCookie` was not validating Terra user enablement/ToS. We previously fixed a similar issue in the Leo-proxy implementation of `setCookie`.

### Testing strategy
- ✅  Unit test coverage of setCookie for enabled/disabled users
- (TODO) test on a BEE

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/ListenerConnectionHandler.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/ListenerConnectionHandler.java
@@ -43,17 +43,13 @@ public class ListenerConnectionHandler {
   }
 
   public boolean isSetCookie(RelayedHttpListenerRequest listenerRequest) {
-    var isSetCookieRequest =
-        listenerRequest.getHttpMethod().equals("GET")
-            && Utils.isSetCookiePath(listenerRequest.getUri());
-    return isSetCookieRequest;
+    return listenerRequest.getHttpMethod().equals("GET")
+        && Utils.isSetCookiePath(listenerRequest.getUri());
   }
 
   public boolean isStatus(RelayedHttpListenerRequest listenerRequest) {
-    var isStatusRequest =
-        listenerRequest.getHttpMethod().equals("GET")
-            && Utils.isStatusPath(listenerRequest.getUri());
-    return isStatusRequest;
+    return listenerRequest.getHttpMethod().equals("GET")
+        && Utils.isStatusPath(listenerRequest.getUri());
   }
 
   public boolean isRelayedWebSocketUpgradeRequestAcceptedByInspectors(
--- a/service/src/main/java/org/broadinstitute/listener/relay/http/ListenerConnectionHandler.java
+++ b/service/src/main/java/org/broadinstitute/listener/relay/http/ListenerConnectionHandler.java
@@ -37,23 +37,23 @@ public class ListenerConnectionHandler {
     return this.inspectorsProcessor.isRelayedHttpRequestAccepted(listenerRequest);
   }
 
-  public boolean isNotPreflight(RelayedHttpListenerRequest listenerRequest) {
+  public boolean isPreflight(RelayedHttpListenerRequest listenerRequest) {
     // TODO: security enhancements, validate origin is valid
-    return !listenerRequest.getHttpMethod().equals("OPTIONS");
+    return listenerRequest.getHttpMethod().equals("OPTIONS");
   }
 
-  public boolean isNotSetCookie(RelayedHttpListenerRequest listenerRequest) {
+  public boolean isSetCookie(RelayedHttpListenerRequest listenerRequest) {
     var isSetCookieRequest =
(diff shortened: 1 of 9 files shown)


service/src/main/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessor.java
```diff
@@ -157,51 +163,59 @@ public Result writeSetCookieResponse(RelayedHttpListenerContext context) {
...
+          String.format(
+              "Origin %s not allowed. Error Code: RHRP-002",
+              requestHeaders.getOrDefault("Origin", "")));
+      return Result.FAILURE;
+    }
 
-        var oauthInfo = tokenChecker.getOauthInfo(authToken.get());
+    try {
+      // Check Sam enablement
+      var enabled = samResourceClient.isUserEnabled(authToken.get());
```
rtitle, line 185: Some refactoring in this method, but the Sam check is the only new thing.

service/src/main/java/org/broadinstitute/listener/relay/inspectors/SamResourceClient.java
```diff
@@ -1,34 +1,39 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.cache.annotation.Cacheable;
 
 public class SamResourceClient {
+  private final String samUrl;
   private final String samResourceId;
   private final String samResourceType;
   private final TokenChecker tokenChecker;
-  private final ApiClient samClient;
```
rtitle, line 16: I was thinking reusing an `ApiClient` could cause thread safety issues since we're calling `setAccessToken` for every request. I changed it to instantiate an `ApiClient` for each request, but use a common thread pool. [WSM](https://github.com/DataBiosphere/terra-workspace-manager/blob/main/service/src/main/java/bio/terra/workspace/service/iam/SamService.java#L100) and [BPM](https://github.com/DataBiosphere/terra-billing-profile-manager/blob/main/service/src/main/java/bio/terra/profile/service/iam/SamService.java#L449) work the same way.

service/src/test/java/org/broadinstitute/listener/relay/http/RelayedHttpRequestProcessorTest.java
```diff
@@ -272,6 +278,66 @@ void writeStatusReponse_notOk() throws IOException {
     }
   }
 
+  @Test
```
rtitle, line 281: Didn't seem like there were existing tests for `setCookie` so I added some.

service/src/main/java/org/broadinstitute/listener/relay/transport/RelayedRequestPipeline.java
```diff
@@ -80,12 +80,18 @@ public void registerHttpExecutionPipeline(Scheduler scheduler) {
     listenerConnectionHandler
         .receiveRelayedHttpRequests()
         .publishOn(scheduler)
-        .filter(c -> listenerConnectionHandler.isNotPreflight(c.getRequest()))
```
rtitle, line 83: The sequence of negative checks were getting a little confusing, so changed it to a `handle` call.

rtitle:
I finally tested this on a BEE and it is working I think. Tested with an enabled user, then disabled and verified `setCookie` no longer works. Requesting some reviews.

service/src/main/java/org/broadinstitute/listener/relay/http/ListenerConnectionHandler.java
```diff
@@ -37,23 +37,19 @@ public boolean isRelayedHttpRequestAcceptedByInspectors(
     return this.inspectorsProcessor.isRelayedHttpRequestAccepted(listenerRequest);
   }
 
-  public boolean isNotPreflight(RelayedHttpListenerRequest listenerRequest) {
+  public boolean isPreflight(RelayedHttpListenerRequest listenerRequest) {
```
LizBaldo, line 40: Death to double negatives! Thanks 😅

service/src/main/java/org/broadinstitute/listener/relay/transport/RelayedRequestPipeline.java
rtitle: I think the `if-statement-in-handle` is equivalent to a chain of `filter..doOnDiscard` in reactor-core. Just felt easier to read to me.

The added Sam call in `/setCookie` could impact performance… but we do it in Leo too, it's cached, and from the UI the `setCookie` call still felt instantaneous to me. 🤷

PR #52, opened 2024-01-04
[IA-4756] update to latest terra common library

LizBaldo:
### What
- It was not too too bad in the end, I just had to publish a new TCL with this fix https://github.com/DataBiosphere/terra-common-lib/pull/129

### Why
- Upgrade to latest TCL that provides a secure version of logback-core

### Testing strategy
- [ ]


cpate4 (approved):
compiles and "tests" pass - :shipit:

PR #54, opened 2024-01-09
build(IA-4442): enforce spotless check in CI

cpate4:
### What
- enforce spotless check in CI, and
- format code when development mode

### Why
- enforces consistency with code styling 

### Testing strategy
- [x] validated that `./gradlew javaCompile` will format the code locally, and
- [x] checked in invalid code on this branch to see build fail

### DIFF
--- a/service/tooling.gradle
+++ b/service/tooling.gradle
@@ -20,7 +20,9 @@ tasks {
 
 	compileJava {
 		options.compilerArgs.add("-Aimmutables.gradle.incremental")
-		// https://broadworkbench.atlassian.net/browse/IA-4442
+
+		// enforce consistency in CI environment, and
+		// enable auto-format in development environment
 		if (isCiServer) {
 			dependsOn(spotlessCheck)
 		} else {
--- a/service/tooling.gradle
+++ b/service/tooling.gradle
@@ -12,8 +12,6 @@ spotless {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


service/tooling.gradle
```diff
@@ -18,7 +18,15 @@ idea.module.downloadJavadoc = true
 
 // Immutables has incremental compilation disabled by default
 tasks {
+	boolean isCiServer = System.getenv().containsKey("CI")
+
 	compileJava {
 		options.compilerArgs.add("-Aimmutables.gradle.incremental")
+		// https://broadworkbench.atlassian.net/browse/IA-4442
```
LizBaldo, line 23: I would not link the ticket, but instead would leave a comment to explain that we want the build to fail on CI if the code is not formatted properly. We can always find the original ticket by looking into the git history since you named the branch with the ticket number

LizBaldo (approved):
Small nit comment but otherwise lgtm! Glad the pseudocode in the ticket was useful

service/tooling.gradle
cpate4: ok

cpate4:
... here is [the failing CI-build](https://github.com/DataBiosphere/terra-azure-relay-listeners/actions/runs/7462243245/job/20304115345?pr=54)!

PR #56, opened 2024-01-19
AJ-1550: logging and health check for HybridConnectionListener

davidangb:
## Summary of changes:
Adds a new `hybridConnectionListenerHealth` health check, via Actuator, to the k8s liveness probe. This health check echoes the state of `HybridConnectionListener.isOnline()`, which is part of the `azure-relay` library.

### Why
It seems that if the listener runs into certain errors - specifically, we've seen problems with `java.util.concurrent.TimeoutException: DNS timeout 15000 ms` - the listener's connections hang and stop responding.

This PR adds the listener's online state to the actuator health check. In theory, this would allow k8s to kill the listener pod and restart a new pod. This is drastic, but it could kick the listener connection back to a good place.

-

### Testing strategy
- [x] Added a unit test that verifies `HybridConnectionListener.isOnline()` affects the liveness probe
- [ ] Needs manual testing, in a real AKS environment, to see if `HybridConnectionListener.isOnline()` is reliable and goes down in the use cases we are seeing. See [this comment](https://github.com/DataBiosphere/terra-azure-relay-listeners/pull/56#pullrequestreview-1836825305)

### DIFF
--- a/service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
+++ b/service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
@@ -19,7 +19,6 @@ import org.broadinstitute.listener.relay.inspectors.SetDateAccessedInspectorOpti
 import org.broadinstitute.listener.relay.inspectors.TokenChecker;
 import org.broadinstitute.listener.relay.transport.DefaultTargetResolver;
 import org.broadinstitute.listener.relay.transport.TargetResolver;
-import org.broadinstitute.listener.relay.wss.WebSocketConnectionsRelayerService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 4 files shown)


service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
```diff
@@ -77,14 +79,23 @@ public RelayedHttpRequestProcessor relayedHttpRequestProcessor(
...
-        tokenProvider);
+    HybridConnectionListener hybridConnectionListener =
+        new HybridConnectionListener(
+            new URI(connectionParams.getEndpoint().toString() + connectionParams.getEntityPath()),
+            tokenProvider);
+
+    hybridConnectionListener.setOfflineHandler(
+        t -> logger.warn("HybridConnectionListener is now offline"));
+    hybridConnectionListener.setOnlineHandler(
+        () -> logger.warn("HybridConnectionListener is now online"));
```
cjllanwarne, line 96: Maybe also have some custom counter metric emitted?

rtitle (commented):
This looks convincing to me. Is there a good way to test it? Maybe we could simulate the case by deleting the hybrid connection for a valid listener, and verify the health check fails and k8s restarts the pod..

service/src/main/resources/application.yml
```diff
@@ -98,5 +98,8 @@ management:
     health:
       enabled: true
       probes.enabled: true
+      group:
+        liveness:
+          include: livenessState,hybridConnectionListenerHealth
```
davidangb, line 103: the liveness probe currently returns the abbreviated `{"status": "UP"}` or `{"status": "DOWN"}`. We could add `show-details: ALWAYS` here so it shows the individual statuses of each of the `livenessState` and `hybridConnectionListenerHealth` subsytems.

service/src/main/java/org/broadinstitute/listener/config/AppConfiguration.java
davidangb: I like this idea! It definitely blows up scope of this PR, though, since prometheus and metrics are not currently enabled in this codebase. If we wanted to enable prometheus/metrics, I'd suggest blowing up scope even further and upgrade to Spring Boot 3 first, since a bunch of metrics stuff changed between 2 and 3.

yuliadub (approved):

rtitle (approved):

service/src/test/java/org/broadinstitute/listener/relay/http/AvailabilityTest.java
```diff
@@ -71,4 +87,44 @@ void livenessState() throws Exception {
...
+   * @throws Exception on exception in the test
+   */
+  @Test
+  void hybridConnectionListenerHealth() throws Exception {
+    // livenessState and liveness probe should both be ok, to start
+    assertThat(
+        "Liveness state should be CORRECT",
+        applicationAvailability.getLivenessState(),
+        equalTo(LivenessState.CORRECT));
+    ResultActions result = mvc.perform(get("/actuator/health/liveness"));
```
jladieu, line 104: Is `result` leftover/redundant code?  It looks identical to `livenessResult` but with a less specific name.

